PR #279, opened 2026-07-30
Spec and plan the store import layering refactor

charlesmsiegel:
Design docs only — no source changes to `backend/src`. Adds a spec, a 17-task implementation plan, and a validator script for giving `backend/src/grimoire/store/` an acyclic module graph with every import at module scope.

## The measurement that changed the problem

`docs/codemap.html` reports a 63-file SCC inside the store. That number is an artifact: the atlas attributes `from . import scenes` to the *package*, so `__init__.py` importing 59 submodules while each submodule imports siblings collapses the package into one node.

Resolved to actual submodules:

- The module-level graph has **one** true cycle today: `scenes → scene_refs → audit → scenes`.
- **53 grimoire-internal imports sit inside function bodies.** Hoisting all of them yields a single **15-module SCC**.
- 4 more in-function imports are third-party (`jinja2` ×2, `tiktoken`, `claude_agent_sdk`), deliberate rather than cycle-driven.

## The seam

Measured attribute usage shows every back-edge is a **lifecycle function reaching sideways** while every forward edge wants only **record-access primitives**:

```
modules   → campaigns : campaign_meta_path, read_campaign      (deferred)
campaigns → modules   : pack_root                              (deferred)
worlds    → campaigns : world_refs                             (deferred)
campaigns → worlds    : world_root, world_exists, canonical_id
```

Nothing outside `campaigns.py` imports `create_campaign`; nothing outside `worlds.py` imports `delete_world`. Splitting each record kind into a leaf *core* and a *lifecycle* layer above it removes all 15 nodes from the SCC without adding indirection. Ten modules become subpackages (63 → ~109 store modules; largest file 1297 → ~250 lines).

## Hazards a file graph does not show

Each was reproduced on a prototype before being written into the spec:

- **Package `__init__` re-exports relocate the fragility rather than removing it.** Reordering three lines of a package `__init__.py` — file graph still fully acyclic — produces `ImportError: cannot import name 'world_refs' from partially initialized module`.
- **A submodule may not share a name with a public function of the same module.** `sheets.read`, `sheets.write`, `sheets.coverage`, `sheets.advance`, `module_edit.rename` and `absorb.materialize` are all public functions; a same-named submodule is overwritten by the facade re-export, giving `AttributeError: 'function' object has no attribute 'list_refs'` at *call* time, past both import and the cycle guard.
- **A cycle can leave a package and come back.** `repad` in `scenes/serialize.py` calls `scene_refs.repoint`, `scene_refs` imports `chronicle`, and `chronicle` needs `scenes.serialize` — clean in every per-package check, a three-package loop overall.
- **Hoisting must preserve an import's form and its fallback breadth.** `jinja2` and `claude_agent_sdk` are used by name, not as modules, so `import jinja2` leaves `Environment` unbound. And `count_tokens` already wraps `_encoder()` in a broad `except Exception`, so narrowing to `except ImportError` while hoisting would let a broken-but-installed `tiktoken` break the store facade at import instead of degrading to the heuristic.

## The tooling is verified, not sketched

**`backend/tests/test_import_guard.py`** (plan Task 2). Five rules: no import inside a function *or class* body; no module-level cycles (self-imports included); no name bound off a package `__init__`; no function bound by value from a cross-package leaf module, wildcards included; and no submodule shadowed by a facade export. It reports exactly 58 violations today (57 deferred + 1 cycle), matching an independent AST count, and every rule was verified both to fire on an injected violation and to produce no false positives on current code.

It uses a **ratchet baseline**: a violation nobody recorded fails, and a recorded line that no longer occurs *also* fails, so the file cannot decay into a standing exemption list. All 58 lines are assigned to exactly one task — checked programmatically, zero unassigned, zero duplicated.

**`scripts/validate_partition.py`** exists because hand-deriving import graphs failed three review rounds running. Four modes, all run by Task 0 and again before each split:

| Mode | Checks |
|---|---|
| *(default)* | intra-package cycles, plus every top-level name having a home — annotated assignments included. An unplaced name is a hard failure, since where it lands can close a loop the graph never saw. A missing spec section fails rather than silently skipping a package. |
| `--graph` | the post-refactor graph **across** packages, resolving import aliases to real module targets so it keeps working as the splits land |
| `--imports` | every cross-package reference resolved through the placement tables, emitting package-qualified aliases. Task 13 generates its import list this way rather than carrying a hand-written table |
| `--plan` | the plan's task interfaces against the spec's placement tables — the drift that produced four separate findings |

All ten packages validate clean in every mode.

## Files

- `docs/superpowers/specs/2026-07-30-store-import-layering-design.md`
- `docs/superpowers/plans/2026-07-30-store-import-layering.md`
- `scripts/validate_partition.py`

## Verification

Backend baseline after the plan's Task 1 test repairs: **2713 passed, 1 failed**. The single failure (`test_atomic.py::test_a_read_only_record_is_not_silently_replaced`) is a container artifact — this environment runs as uid 0, so chmod 0444 cannot stop a write. Green on a normal user account.

Task 1 repairs four broken `stat` stubs, none of which are part of this diff: three in `test_atomic.py` take one positional arg while Python 3.11.15's `pathlib` passes `follow_symlinks=`, aborting pytest's teardown with an INTERNALERROR; and one in `test_assets_store.py` raises for every path but `avatar.png` — including the directory `image_path` stats first — so the scan it means to exercise never runs on any uid.

## Review status

`CLAUDE.md` requires codex gates at each stage. `codex` is not installed in this container, so the spec→plan gate was substituted with a manual adversarial pass. The repo's GitHub codex integration has since reviewed the PR across **thirteen rounds: 53 findings, none spurious, all addressed** — see the summary comments below.

Four defect classes are now checked mechanically rather than by review: intra-package cycles, cross-package cycles, cross-package import lists, and spec/plan placement agreement. The most recent rounds have found bugs in those checks rather than in the design; the spec's structure has been stable since round five.


charlesmsiegel:
All nine findings verified against the code before acting. None were spurious; all are fixed in `d3dabdb`. One comment rather than nine thread replies, since the changes are interlocking.

### Three changed the structure

**Repartition helpers that call back across proposed files (P1).** Confirmed all four cases: `world_root_of` calls `read_campaign` which calls `campaign_meta_path`; `load_pack_at` calls `_load_rules`/`_load_content` while `read_content` calls `pack_root`; `_lock` calls five version helpers; `_serialize_messages` calls `_append_block` while `append_message` calls `_block`. The tables were derived from what each function is *about* rather than what it *calls*. Placement now follows the call graph in both spec and plan, with the rule stated explicitly: sketch the intra-package call graph first, and if two proposed files call into each other, merge them or pull the shared helper into a third.

**Retarget campaign consumers (P1).** Confirmed — `appearances.py:16` and `sheets.py:18` among ten modules importing the `campaigns` facade. Each split task now retargets its consumers in the same commit, with a `grep` recipe to find them; a missed one surfaces as a reported cycle in that task's own verification.

**Patch the bindings split-module callers actually use (P1).** Confirmed and the plan was wrong in the more damaging direction. `routes/scenes.py:299` calls `store.audit.materialize(...)` — a call-time lookup on the package — so the existing facade patch keeps working and my retarget would have silently stopped injecting the crash. Same for `test_locks_store.py:865`, which drives an HTTP route reaching `store.scenes.append_message` via `routes/mechanics.py:40`. The rule is now caller-driven, with a per-site table: four patches stay on the facade, and each remaining retarget moves to the task that rewrites *that caller's* import — `modules.resolve` to Task 7, `modules.load_pack` to Task 8, `campaigns.read_campaign` to Task 13. Every retarget now carries a step proving the patch still injects.

### Three were guard bugs, each reproduced then re-verified

**Marker siblings (P2).** Reproduced: with `# import-ok:` applied to a real deferred import, the pre-fix guard still reported the violation. `others` included the node itself, so `_shares_first_line(node, node)` made `_sole_owner` reject every inline marker — the documented escape hatch could never fire. Fixed with identity exclusion; the marker now suppresses, and Task 2 gained a step that proves it.

**Singleton SCCs with self-imports (P2).** Fixed and verified: `_cycles({'a': {'a'}})` now reports `['cycle a']` while `{'a': {'b'}}` stays empty. Task 2 asserts both.

**Absolute package imports (P2).** Fixed and verified — an injected `from grimoire.store.calendars import CalendarError` is now caught as `form grimoire.store.suggest::grimoire.store.calendars.CalendarError`. Confirmed the guard still reports exactly 58 on clean code afterwards, so no false positives.

### Three more

**Vacuous public-API check (P2).** It compared `__all__` with itself. Now diffs against a snapshot captured in Task 2 covering `dir()` as well as `__all__` — necessary because `module_display` is bound only as a side effect of `modules.py:18` and never appears in `__all__`, so an `__all__`-only snapshot would not notice it vanishing. Captured: 81 exported, 101 public attributes.

**Claude SDK test doubles (P2).** Confirmed — `install_fake_sdk` swaps `sys.modules["claude_agent_sdk"]`, which a module-level import has already read past. Task 14 now patches the `grimoire.claude_agent.claude_agent_sdk` global instead, adapts the missing-SDK test at line 82, and proves the fake is observed by removing the patch and requiring failure.

**The asset test is not a root artifact (P2).** Correct, and this one invalidated a stated baseline. The trace is `FileNotFoundError` on the *directory* `.../assets/default`, exactly as described — `image_path` opens with `if not d.exists()`. Narrowing the stub to `avatar.jpg` makes it pass. Measured baseline after both repairs: **2713 passed, 1 failed**, the remainder being the genuinely chmod-dependent `test_a_read_only_record_is_not_silently_replaced`.

---

charlesmsiegel:
Second round: all seven verified in the source, all fixed in `711726d`. Baseline coverage re-checked — 58 live, 58 assigned, zero unassigned, zero duplicated.

### Three were latent cycles inside a single package

**`journal ↔ migrate` (P1).** Confirmed at `module_edit.py:268` (`recover` → `_campaign_locks`), `:318` (`_replay_journal` → `_run_migration`), `:565-566` (`_apply` → `recover`, `_require_user_root`). There is no `journal.py` now — crash recovery and migration are one concern, and the journal being replayed *is* the migration journal.

**`world_state ↔ assemble` (P1).** Confirmed: `activate` at `context.py:24`, `_world_info` calls it at `:200`, `_assemble` calls `_world_info` at `:590`. `activate` moves in beside `_world_info`, so the edge runs one way.

**`repoint_scenes` (P1).** Confirmed: `scene_refs.py:14` imports `appearances` and iterates the six ref-holding stores at `:21`. In `transitions.py` it would have closed `scenes → scene_refs → appearances.transitions → scenes`. It calls only `record` and `_write`, so the record-level leaf is where the call graph puts it regardless.

### The naming collision is worse than reported, and fails at call time

Not two collisions but **six**: `sheets.read`, `sheets.write`, `sheets.coverage`, `sheets.advance`, `module_edit.rename`, `absorb.materialize` are all existing public functions. Reproduced on a prototype:

```
AttributeError: 'function' object has no attribute 'list_refs'
```

The failure is at *call* time, after passing both import and the cycle guard — the facade re-export overwrites the package attribute that held the submodule. Files renamed to `reader.py`, `writer.py`, `tally.py`, `advancement.py`, `renaming.py`, `materializer.py`; every facade function name is unchanged.

Renaming alone would only fix the six known cases, so there's now a guard rule for the class — `test_no_submodule_is_shadowed_by_a_facade_export` asserts every package attribute matching a submodule filename is still a module. Verified passing against the current package and detecting both collisions in the prototype.

### Three more

**`DEFAULT_BUILTIN_DIR` (P1).** Confirmed at `modules.py:39`. Relocated to `modules/pack.py` it would resolve to `store/modules/builtin_modules/` and `pack_root` would raise `ModuleNotFound` for every shipped pack. The expression walks one level further up; the data directory stays put for Android packaging. A grep for `__file__` across the store confirms this is the only affected site — `epub.py:26` and `climates/__init__.py:20` compute paths the same way, but neither module moves. Task 5 now verifies `builtin_dir()` lists `['d20-basic', 'pool-basic']`.

**Cycle baseline timing (P1).** Correct — retargeting `scene_refs` to `audit.baselines` clears the SCC in Task 8, so holding the line until Task 9 would fail Task 8's own stale-entry check. Moved; re-verified that the line is now assigned to task 8 and to nothing else.

**`module_display` in `__all__` (P2).** Correct — Task 2 freezes a list that has never contained it, so adding it would fail the very check Task 5 runs. The alias alone restores both `from grimoire.store import module_display` and the `dir()` entry, which is what the captured snapshot covers.

---

charlesmsiegel:
Third round: all four verified in the source, fixed in `f4ebd79`. Coverage re-checked — 58 live, 58 assigned, zero unassigned, zero duplicated.

### Two more single-package cycles

**`_split_csv` (P1).** Confirmed: defined at `modules.py:388`, called from `_load_rules` at `:407`/`:410` — which moved to `pack.py` — while `read_content`/`read_rule` call `pack_root` at `:476`/`:710`. It follows the loaders into `pack.py`; `content.py` keeps only `read_content` and `read_rule`.

**`layout ↔ edits` (P1).** Confirmed: `_prune_layout` uses `_read_json`/`_write_json` at `:771`/`:792`, while `upsert_group`/`delete_group` call `_prune_layout` at `:1142`/`:1155`. Those accessors have eight call sites across the package, so they become a `packfile.py` leaf (`_read_json`, `_write_json`, `_read_sheets`) that both files import rather than being co-located with either.

### A contradiction I introduced

**`scenes.append_message` (P1).** Correct — Task 9 still said to retarget it, contradicting the caller-based table added in the previous round. The test drives an HTTP route, `routes/mechanics.py:40`/`:202` read the facade at call time, so the retarget would have recorded only the roll write and failed the assertion at `:869`. The table was right; the task was stale text I should have removed when I wrote the table.

### Hoisting must preserve the import's form — and this was the fourth instance

**SDK member references (P1).** Confirmed at `claude_agent.py:35-36`: `stream()` pulls six names out of the SDK, so a bare `import claude_agent_sdk` leaves `ClaudeAgentOptions` unbound and raises `NameError` before `query` is reached. The sentinel now covers each name individually. Two of them are `except` targets, so they get empty tuples rather than `None` — verified: `except None` raises `TypeError: catching classes that do not inherit from BaseException is not allowed`, while `except ()` simply never matches.

This also invalidated the fix from round two, which patched a `claude_agent.claude_agent_sdk` global that this form does not create. `install_fake_sdk` now patches all six names, and the missing-SDK test sets the `query` sentinel.

Checking the neighbouring steps for the same mistake found a fourth instance Codex hadn't flagged: `prompts.py:30` and `epub.py:32` both do `from jinja2 import Environment, FileSystemLoader, StrictUndefined` and call those names bare, so my instruction to hoist `import jinja2` would have produced the identical `NameError`. Fixed, and the rule is now a global constraint — check what the body references before moving the line. `tiktoken` is the only one of the four genuinely used as a module.

### One more, found while checking the above

Task 13's hoists are not verbatim moves. `from . import appearances` binds a *package* after Tasks 3-9, closing `chronicle → appearances → transitions → scenes → scene_refs → chronicle`. Each row now targets a submodule and rewrites its call sites with it; `calendars/base.py` is the one exception, since `plugins` is a plain module and carries no partial-initialization hazard.

---

charlesmsiegel:
Fourth round, all three fixed in `22b9f4c` — but the partition finding was the third of its kind, so I stopped hand-fixing and built the check instead.

### `scripts/validate_partition.py`

It parses the partition tables out of the spec, resolves every top-level name in the source module to its proposed file, builds the file-level call graph, and reports SCCs with the function pairs that form them. It runs as Task 0 and again before each split.

Its first run answered this round's finding and went further: `module_edit`'s tangle is **four** files, not the two reported —

```
CYCLE: edits <-> layout <-> migrate <-> renaming
  edits    -> renaming via delete_group()->_field_keys, upsert_group()->_group_scope
  layout   -> renaming via _specialize_layout()->_fragment_users
  migrate  -> renaming via _apply()->_RenameCollision
  renaming -> edits    via rename()->check_proposal_guard
```

The back-edges are all small shared helpers, so `_RenameCollision`, `_field_keys`, `_group_scope` and `_fragment_users` become a `scope.py` leaf, and `check_proposal_guard` moves to `renaming.py` (`rename` at `:939` is its only in-package caller). All ten packages now report `[OK ]`.

The other nine were already clean — established in one pass rather than one review round each, which is the point.

### The guard did not enforce this plan's own rule (P2)

Correct, and it's the more embarrassing of the two. `from ..modules.binding import resolve` targets a leaf module, so my `_check_form` skipped it entirely — while it still binds by value and silently defeats any test patching `modules.binding.resolve`. Rule 3 now covers cross-package leaf modules.

Narrowed to **functions** after the first version produced a false positive on `weather/seasons.py` importing the `CalendarProvider` *class*: a class or exception bound by value carries neither the init-order nor the patchability hazard. Verified both directions — still exactly 58 violations on current code, and an injected `from ..calendars.base import get_provider` is caught as `form grimoire.store.weather.seasons::grimoire.store.calendars.base.get_provider`.

### `packfile.py` in Task 10's create list (P2)

Correct — the spec assigned it, the task never named it. Both `packfile.py` and the new `scope.py` are in the create list now, with the reason each exists spelled out.

---

Four rounds, 23 findings, none spurious. The recurring one was always the same shape — a table that reads well but implies a cycle — and it kept recurring because I kept re-deriving the graph by hand. That should now be settled mechanically rather than by the next round of review.

---

charlesmsiegel:
Fifth round, both verified and fixed in `05e71f7`. The first one lands on the tool I added last round specifically to settle this class of defect — it didn't.

### The validator was unsound (P1)

Confirmed on both counts. `toplevel_names()` handled `Assign` but not `AnnAssign`, so `FILE_KINDS: tuple[str, ...] = ...` in `sheets.py` was invisible despite four proposed files sharing it. And every unassigned name collapsed into one synthetic node, so `[OK]` meant "acyclic among the names I happened to enumerate" — not what I claimed for it.

`SYNTHETIC_SPEAKERS` is the case that bites, exactly as described: read by both `turns.py` and `write.py`, so placing it in `write.py` closes `turns → write` while write functions already depend on turn helpers. The old tool exited 0 on that.

Three changes: it reads annotated assignments; each unplaced name gets its own node so one cannot mask another; and an unplaced name is now a hard failure that names the files reading it.

Re-running it turned up **55 unplaced names** — including the entire contents of `module_display.py`, which the spec described in prose (*"all of today's `module_display.py`"*) rather than enumerating, so the checker never saw `_load_theme`, `_Expander`, `_type_scope` or the other sixteen. All 55 are now placed with their readers, and all ten packages report:

```
[OK   ] campaigns: 21 names mapped, 0 unassigned
...
[OK   ] context: 39 names mapped, 0 unassigned
```

which for the first time means what it says.

### Dependency-module patch targets (P1)

Confirmed all three. `sheets.atomic`, `audit.locks` and `scenes.calendars` are `from . import <dep>` today, so the attribute is the shared `grimoire.store.<dep>` module; after the split it lives on a submodule and the facade `__init__.py` does not re-bind it, so the current spelling raises `AttributeError` before the assertion runs.

They now point at the shared module directly — which is what they always meant, and is immune to further reshuffling:

```python
from grimoire.store import calendars
    monkeypatch.setattr(calendars, name, watched)   # test_locks_store.py:193
```

Assigned to Tasks 7, 8 and 9 respectively, alongside those modules' import rewrites. `test_locks_store.py:196` already carries its own injection proof (`assert free, "the calendar was never resolved -- the test proves nothing"`); the other two get the standard drop-the-patch check.

I scanned the whole suite for `<split module>.<dependency>` rather than taking the three as given — there is no fourth.

---

Five rounds, 25 findings, none spurious. Notable that this round's headline finding was a defect in last round's fix: the check I introduced to stop hand-derivation from failing was itself hand-derived, and wrong in the same way. It's now checked against the property it claims.

---

charlesmsiegel:
Sixth round, all four verified and fixed in `21380a9`. One of them I fixed twice — the first attempt was wrong and I backed it out.

### Class-body imports (P2)

Confirmed by injection: `class _Probe: import json` produced no violation at all. Only `FunctionDef` pushed the scope stack, so a class body counted as module scope — in a check whose entire claim is that every import is at module scope.

Fixed. It also qualifies method scopes now, which renamed one baseline entry (`stream` → `ClaudeAgentClient.stream`, since it is a method of `ClaudeAgentClient`). The plan's list is updated and coverage is back to **58 live, 58 assigned, 0 unassigned, 0 duplicated**.

### `world_root_of` in Task 3 (P2)

Confirmed — Task 3's interface block still put it in `paths.py` after the spec and the plan's own placement table moved it to `read.py`. Stale text I failed to update when I made that correction. Fixed, with the reason inline so it does not drift back.

### Response-presets patch in Task 13 (P1)

Confirmed. Task 13 rewrites `usage()` to hold `campaigns_read`, which is exactly what finally moves that patch — and neither the file list nor the steps touched `test_response_presets.py:601`. The injected `OSError` would never fire and the `unevaluated` row would go missing. Added as Step 1b with the injection proof.

### `_campaign_locks` binding (P2) — and an over-broad fix I backed out

The concern is real: `delete_module` lives in `packs.py` while `_campaign_locks` lives in `migrate.py`, so a `from .migrate import _campaign_locks` would cache the original and leave `test_module_edit.py:914` patching something nothing calls — the ordering assertion never reached.

My first fix was a blanket guard rule rejecting sibling name-imports inside split packages. **I backed it out.** It produced six false positives across `calendars/` and `weather/`, and two of them were `from .plugins import load_custom_providers` — the exact line Task 13 instructs the implementer to write. A rule that contradicts the plan it is meant to enforce is worse than no rule.

The operative rule is narrower and now stated where it belongs: *any function a test patches is reached through a module object by its caller*, in or out of package. The enumerated patch table is the list, and the drop-the-patch injection proof is the check. Task 10 now states that `packs.py` must call `migrate._campaign_locks(...)` through the module object.

---

Six rounds, 29 findings, none spurious. Worth noting what changed this round: checking for false positives before shipping the guard rule is what caught my own bad fix, rather than the next review round catching it. That check exists because rounds 4 and 5 didn't have it.

---

charlesmsiegel:
Seventh round, all three verified and fixed in `1da4eda`. The tiktoken one was half a finding — the same argument applies to a second site that wasn't flagged.

### Wildcard imports (P2)

Confirmed. `_kind_of(target, "*")` returns `other`, so `from ..x.y import *` walked straight through the function test the leaf-module rule turns on — binding every exported function by value, which is precisely what that rule exists to stop.

Fixed and verified in both directions: an injected `from ..calendars.config import *` is now reported as `form grimoire.store.weather.seasons::grimoire.store.calendars.config.*`, and clean code still measures exactly 58.

(The package branch already caught `*` — `f"{target}.*" not in MODULES` is trivially true there. Only the leaf branch had the hole.)

### `_lock` in Task 6 (P2)

Confirmed — the same stale-text failure as `world_root_of` last round, in a task interface rather than the placement table. Fixed, with the reason inline. Task 6's interface block now also spells out `versions.py`, which it had omitted entirely.

### tiktoken fallback breadth (P1 in substance)

Confirmed and this is the one that would have bitten hardest. `count_tokens` wraps `_encoder()` in a broad `except Exception` (`context.py:726-729`), so an installed-but-broken tiktoken — an incompatible build raising `ValueError` or `OSError` at import — degrades to the heuristic today. Hoisting behind `except ImportError` would instead let that escape a *module-level* import and break the whole store facade. Turning a silent fallback into an app that will not start is not a behaviour-preserving refactor.

Now `except Exception`, with the reasoning recorded so it does not get "tidied" back later.

**The same argument applies to `claude_agent_sdk`, which wasn't flagged.** `llm.py:11` does `from .claude_agent import ClaudeAgentClient` at module scope, so a broken SDK that today fails one request would, after hoisting with an `ImportError`-only guard, stop the app from starting. Both sites now catch broadly. One deliberate difference is recorded: for an installed-but-broken SDK the error becomes `missing_dependency` at call time rather than the SDK's own exception. The missing case — the one that actually happens on Android — is unchanged.

---

Seven rounds, 32 findings, none spurious. Validator clean across all ten packages; guard baseline 58 live / 58 assigned / 0 unassigned / 0 duplicated.

---

charlesmsiegel:
Eighth round, all six verified and fixed in `fd64b61`. Two of them made me stop hand-writing another table.

### Task 13's import list was materially short (P1)

Confirmed, and worse than reported once I computed it rather than checked the three named cases. Every gap is a runtime `NameError`/`AttributeError` that the cycle guard cannot see by construction.

So the validator grew an `--imports` mode that resolves every cross-package reference through the spec's placement tables and prints the exact submodules each consumer needs:

```
--- chronicle.py
    from .campaigns import paths
        campaigns.campaign_root -> paths
    from .scenes import read, serialize
        scenes.TRANSITION_SPEAKER -> serialize
        scenes.get_location_history -> read
        scenes.get_time_history -> read
```

Task 13 now says to run it rather than follow a table. It exits non-zero if any referenced name has no home, so a clean run doubles as a completeness check.

Getting that right needed one fix first: filtering to names the file actually imports. `_type_scope(sheets: dict, tid: str)` shadows the store module with a parameter, so counting `sheets.get` reported a store function that does not exist — a false positive that would have sent an implementer looking for it.

### Missing spec section exited 0 (P2)

Confirmed and fixed. A renamed heading silently dropped a whole package from the check while Task 0 still reported clean — the same "OK means less than it claims" failure the validator had two rounds ago, in a different place. Verified by renaming `### \`worlds/\`` and watching it exit 1 with `parsed 9 of 10 spec sections -- missing: worlds`.

### Three stale-text findings

- **`campaigns/read.py` imports `worlds` too** (P1). Correct, and a direct consequence of moving `world_root_of` there two rounds ago — it calls `worlds.world_root` and catches `worlds.WorldNotFound`. Left on the package facade it closes `campaigns.read → worlds → worlds.lifecycle → campaigns.read` after Task 4. Both Task 3 and Task 4 Step 2 now cover both files.
- **`repoint_scenes`** still listed under `transitions.py` in Task 6's interface while the entry above and the note below put it in `paths.py`.
- **`rename.py`** in Task 10's baseline note, for a file the same task deliberately names `renaming.py` to avoid shadowing the public `rename` function.

### SDK test snippet (P2)

Confirmed — the test imports only `ClaudeAgentClient` and `ClaudeAgentError`, so my `monkeypatch.setattr(claude_agent, ...)` snippet would have raised `NameError` on every `install_fake_sdk` call. Module import added, and `test_claude_agent.py` added to Task 14's file list.

---

Eight rounds, 38 findings, none spurious. Validator clean in both modes; guard baseline 58 live / 58 assigned / 0 unassigned / 0 duplicated.

---

charlesmsiegel:
Ninth round, all four verified and fixed in `ceb421e`. The prose one prompted me to make that whole class mechanical, which is what I'd said I would do rather than hand-patch again.

### `RESPONSE_FIELDS` (P2) — and a new check for its class

Confirmed: Task 9's interface said `serialize.py`, the spec says `write.py` alongside `set_response`, and `campaigns/lifecycle.py` retargets `scenes.RESPONSE_FIELDS` by that path.

This is the fourth instance of one pattern — a correction landing in the spec but not in a task's `Interfaces` block (`world_root_of`, `_lock`, `repoint_scenes`, now this). Each was found only by a human reading two documents side by side. So `validate_partition.py --plan` now compares them directly.

Its first run reported six disagreements. **Five were false positives of my own making** — my explanatory prose after the em dash (`— **not** \`paths.py\`: ... those call \`record\`/\`_ref\`/\`_write\` back`) legitimately names functions living elsewhere, and the regex was reading it as placement. It now reads only the backticked list before the dash, and reports exactly the one real case. Wired into Task 0.

### `--imports` skipped nested packages (P2)

Confirmed and the completeness claim was false. `STORE.glob("*.py")` never saw `weather/__init__.py` — a file Task 13 explicitly modifies for its `scenes.get_location_history` call — so the generator printed nothing for it and exited 0 anyway. Now walks nested packages with the correct relative depth:

```
--- weather/__init__.py
    from ..scenes import read
        scenes.get_location_history -> read
```

That also surfaced `weather/overrides.py`, a consumer nothing had noticed.

### Missing source module exited 0 (P2)

Confirmed — a warning, an empty graph, `[OK]`, exit 0, certifying a partition nothing was read for. Now an error, with a `[UNREAD]` status instead of `[OK]`, exempted only when the corresponding subpackage already exists (a completed split). Verified all three states.

### SDK failure mode (P2)

Right, and better than what I wrote. Collapsing both arms into `missing_dependency` discards the real exception and misreports an installed-but-broken SDK as a missing one. The import error is now captured and re-raised from `stream()`, so a broken install fails at the same call with the same exception, while the module stays importable for `llm.py:11`. Behaviour is preserved in *both* cases rather than one; my earlier draft accepted the misreport as a "small behaviour change" when it was avoidable.

`tiktoken` needs no equivalent — `count_tokens` already swallows everything from `_encoder()`, so a single broad catch there *is* the existing behaviour.

---

Nine rounds, 42 findings, none spurious. Three defect classes are now checked mechanically rather than by review: intra-package cycles, cross-package import lists, and spec/plan placement agreement. All three modes green; guard baseline 58 live / 58 assigned / 0 unassigned / 0 duplicated.

---

charlesmsiegel:
Tenth round, all five verified and fixed in `675fc32`. The `repad` finding exposed a real gap in my tooling, not just in the plan.

### The `repad` cycle (P1) — and the check that should have caught it

Confirmed: `repad` calls `scene_refs.repoint` (`scenes.py:180`), `scene_refs.py:14` imports `chronicle`, and `chronicle` must import `scenes.serialize` for `TRANSITION_SPEAKER`. It also contradicts the spec's own stated invariant that `lifecycle.py` is the only file touching `scene_refs.repoint`.

**Every cycle check I had ran per package**, so a loop that leaves a package and returns was structurally invisible — `scenes` reported clean throughout. `validate_partition.py --graph` now builds the post-refactor graph across the whole store:

```
  CYCLE among: chronicle, scene_refs, scenes.read, scenes.serialize
      chronicle -> scenes.read
      chronicle -> scenes.serialize
      scene_refs -> chronicle
      scenes.read -> scenes.serialize
      scenes.serialize -> scene_refs
```

Getting it right took two corrections of my own before I trusted it. Attribute-only walking missed `scene_refs.py:21`, where the modules are referenced *bare* in a tuple and reached through a loop variable — my first version reported clean. Resolving those to every file in the package then over-approximated into a single 32-node SCC, so they now resolve through the attribute the loop actually calls. And the report joined SCC members with arrows as though they were a path Tarjan had computed.

`repad` moves to `scenes/lifecycle.py`. `--plan` immediately caught the plan-side copy still saying `serialize.py`, which is exactly what that check exists for. Verified by reverting the move and watching `--graph` fail.

### Four smaller ones

- **Missing-SDK test (P2).** Right — patching only `query` leaves `_SDK_IMPORT_ERROR` as `None` whenever the suite runs with the extra installed, so `stream()` hits `raise None` and gives `TypeError` instead of `ClaudeAgentError`. Both names are patched now.
- **Race-window helper (P2).** Confirmed — the snippet used `frontmatter`, `scenes_read`, `scenes_write`, `scenes_moment` and `scenes_lifecycle`, none of which the test binds. Imports added, and `test_scene_store.py` added to Task 9's Modify list.
- **Calendars plugin form (P2).** Confirmed — two adjacent paragraphs prescribed contradictory forms, so following one import with the other's call sites leaves a name undefined. Settled on `from .plugins import load_custom_providers` with call sites unchanged, stated once.
- **Task 0's commit step (P2).** Confirmed — the validator ships with the plan, so `git commit` on an empty index would have stalled the sequence at its first task. Removed.

---

Ten rounds, 47 findings, none spurious. Four defect classes now checked mechanically: intra-package cycles, cross-package cycles, cross-package import lists, and spec/plan placement agreement. All four modes green; guard baseline 58 live / 58 assigned / 0 unassigned / 0 duplicated.

---

charlesmsiegel:
Eleventh round, both verified and fixed in `e7c35a9`. Both are defects in the tooling I added, not the design.

### Unaliased leaf imports collide (P1)

Confirmed with the exact case named. `response_presets.py` needs `read` *and* `paths` from both `campaigns` and `scenes`, so the generator was emitting:

```
    from .campaigns import paths, read
    from .scenes import paths, read
```

The second pair silently rebinds the first. No import error, no guard violation — campaign lookups would just start resolving against scene modules at runtime. An implementer copying the output I advertised as "exact" would have hit it.

Every binding is package-qualified now, and each mapping line prints the literal call-site rewrite rather than just the destination file:

```
    from .campaigns import paths as campaigns_paths, read as campaigns_read
        campaigns.CampaignNotFound -> campaigns_paths.CampaignNotFound
        campaigns.list_campaigns -> campaigns_read.list_campaigns
    from .scenes import paths as scenes_paths, read as scenes_read
        scenes.SceneNotFound -> scenes_paths.SceneNotFound
```

Task 13's sample output is updated to match.

### `--graph` was never actually run (P2)

Confirmed, and this is the more embarrassing of the two. I added `--graph` last round specifically because per-package checks cannot see a cycle that leaves a package and returns — then left it reachable only through a flag Task 0 never invokes. Task 0 ran the default and `--plan` modes; my Step 2 prose said "run all four modes" while the actual command blocks were two.

So the exact defect class `--graph` exists for could have passed every prescribed check, which is how `repad` got as far as it did. `--graph` and `--imports` are now Steps 1c and 1d with their own expected output, not prose.

---

Eleven rounds, 49 findings, none spurious. All four modes green; guard baseline 58 live / 58 assigned / 0 unassigned / 0 duplicated.

Worth noting for the record: both of this round's findings were in tooling added within the last two rounds, and neither was a design problem. The spec's structure has been stable since round five.

---

charlesmsiegel:
Twelfth round, all three verified and fixed in `a5147dc`.

### `--graph` would have gone blind at the splits (P2)

Confirmed, and this is the one that mattered. The walker recorded an import's *local name* only, so the moment a split task writes `from ..worlds import paths as worlds_paths`, `node_for()` would match `worlds_paths` against nothing and drop the edge. The check would have degraded precisely as the packages it exists to check came into being — and reported clean while doing it.

Each binding now resolves to its store-relative target. Verified against all four forms it will actually meet:

```
  worlds_paths   -> worlds.paths        # from ..worlds import paths as worlds_paths
  read           -> campaigns.read      # from ..campaigns import read
  serialize      -> campaigns.serialize # from . import serialize
  baselines      -> audit.baselines     # from grimoire.store.audit import baselines
```

And the `repad` cycle is still caught after the rewrite, so detection did not regress.

### Facade patches in the spec (P2)

Confirmed — the spec carried a blanket "retarget every site" instruction while the plan carries the caller-based exceptions. `routes/mechanics.py:40`,`:202` and `routes/scenes.py:299` reach `store.scenes.append_message` and `store.audit.materialize` at call time, so retargeting to `scenes.write` / `audit.apply` would leave the lock-depth and audit-crash tests passing while injecting nothing. The spec now states the rule as caller-decided and names both exceptions.

### Sequencing skipped the unsplit-module hoist (P2)

Confirmed. The spec went straight from the size splits to the four third-party imports, then claimed the baseline was empty — but the 26 deferred imports in `chronicle`, `proposals`, `response_presets` and nine others are still live at that point, so the final step could not have passed. Added as step 4, with the note that these are not verbatim moves since the targets are packages by then.

---

Twelve rounds, 52 findings, none spurious. All four validator modes green; guard baseline 58 live / 58 assigned / 0 unassigned / 0 duplicated.

The `--plan` check added in round nine covers placement drift between the two documents. The two spec findings here were a different kind of drift — rules and sequencing, not placement — which nothing mechanical covers and I don't think is worth building a checker for.

---

charlesmsiegel:
Thirteenth round, one finding, verified and fixed in `c130845`. It was execution-blocking, which is why it got the full treatment.

### Absorbed sources looked for a package that will never exist (P1)

Confirmed by simulating the post-Task-5 tree — deleting `modules.py` and `module_display.py`, creating `modules/__init__.py`, and running the validator:

```
  !! missing source .../module_display.py and no module_display/ package -- cannot validate this partition
  [UNREAD] modules: 64 names mapped, 0 unassigned
  exit: 1
```

`module_display.py` is absorbed into `modules/display.py`, so there is no `module_display/` package and never will be. The completed-split check looked for one anyway, so the very next validator run — the one Task 6 is required to make — would have failed, blocking every remaining split. The plan would have stalled at its sixth task with a green test suite and a red validator.

The completed replacement is always the *owner* package. Verified both directions: the post-Task-5 tree now reports `[OK    ] modules` and exits 0, and a genuinely missing source with no replacement still fails.

---

Thirteen rounds, 53 findings, none spurious. All four modes green; guard baseline 58 live / 58 assigned / 0 unassigned / 0 duplicated.

### A note on how I will handle further rounds

This one earned a fix: it would have halted execution partway through. Going forward I will triage rather than treat every round as automatically actionable — fixing what would break execution, and noting the rest on the thread without a commit. Twelve of the last thirteen rounds have been document review with no source changed, and the last four have found bugs in the checking tools rather than in the design, which is a signal that the useful information is now in running the plan rather than reading it.

Nothing in this round changes the design. The spec's structure has been stable since round five, and all four mechanical checks pass.

---

charlesmsiegel:
Fourteenth round, all three verified and fixed in `45fd5fd`. All 56 review threads are now resolved.

All three were ways the validator would have grown *weaker* during execution — silently, while still reporting green — which is worse than failing outright, so they cleared the triage bar.

### `--imports` missed local-variable dispatch (P2)

Confirmed. `scene_refs.repoint` iterates `(appearances, audit, ...)` and calls `mod.repoint_scenes(...)`, naming no attribute on the modules themselves — so the file was omitted from the output entirely, despite being exactly the kind of consumer a split has to retarget. `store_graph()` already had the local-attribute resolution for this; `consumer_imports()` didn't. Now:

```
--- scene_refs.py
    from .appearances import paths as appearances_paths
        appearances.repoint_scenes -> appearances_paths.repoint_scenes
    from .audit import baselines as audit_baselines
        audit.repoint_scenes -> audit_baselines.repoint_scenes
```

Which is precisely what was predicted.

### `--graph` recorded phantom module paths (P2)

Confirmed. `from .paths import helper` — the sibling form the plan explicitly permits — was recorded as `campaigns.paths.helper`, so every bare use of `helper` failed the existence check and contributed no edge. An implemented package could report acyclic while missing `campaigns.read → campaigns.paths` entirely. A name imported *from* a module now binds to that module unless the name is itself a module.

### `--graph` stopped at package facades (P2)

Confirmed. Once split, an `__init__.py` is almost all imports and contributes no names, so a consumer edge to `campaigns` terminated at a synthetic node rather than continuing to `campaigns.lifecycle`, `campaigns.read`, `campaigns.paths`. Importing a facade genuinely does run every leaf its `__init__` loads, so package targets now fan out accordingly — which is the only way the package-initialisation cycles staged implementation can create become visible.

Verified no false positives on the current tree in all four modes, and that the `repad` cross-package cycle is still caught after the rewrite.

---

Fourteen rounds, 56 findings, none spurious. All threads resolved. All four validator modes green; guard baseline 58 live / 58 assigned / 0 unassigned / 0 duplicated.

---

PR #281, opened 2026-07-30
Give store/ an acyclic module graph with every import at module scope

charlesmsiegel:
Implements `docs/superpowers/specs/2026-07-30-store-import-layering-design.md` via `docs/superpowers/plans/2026-07-30-store-import-layering.md` (both merged by #279, no implementation had started).

Ten oversized `store/` modules become layered subpackages — a leaf *core* (paths, reads, primitive writes) and a *lifecycle* layer above it that owns create/delete cascades. Every back-edge in the old graph was a lifecycle function reaching sideways, so cutting on that seam removes the cycles without adding indirection.

Rebased onto `3e6d40b`.

## Result

| | before | after |
|---|---|---|
| in-function grimoire imports | 57 | **0** |
| module-level import cycles | 1 (`scenes → scene_refs → audit → scenes`) | **0** |
| largest split-module file | 1297 (`module_edit.py`) | **397** |
| store modules | 63 | 135 |

`store/__init__.py`'s exports and `__all__` are unchanged, so all ~338 `from grimoire.store import x` call sites are untouched. Verified by importing both facades side by side: `__all__` and `dir()` are identical.

## How it is enforced

`backend/tests/test_import_guard.py` parses the package's own ASTs. It fails on any import inside a function or class body, any cycle in the module graph, and — inside `store/` — any cross-package import that binds a **name** off a package rather than a submodule.

That third rule is the non-obvious one. `from ..campaigns import world_refs` raises at import time while `campaigns/__init__.py` is still running, and `from ..campaigns.read import world_refs` binds the function by value so a test patching it silently stops intercepting. **The file graph stays acyclic either way**, so the cycle check alone catches neither. The rule is `from ..campaigns import read` then `read.world_refs()`.

It landed with a 58-line ratchet baseline; each task deleted its own lines, a stale line failed as loudly as a new violation, and the last commit deletes the file and the machinery that reads it. There are now zero `# import-ok:` exemptions in the package.

A second guard, `test_store_api_baseline.py`, freezes the facade against a snapshot covering `dir()` as well as `__all__` — `module_display` was reachable only as a side-effect binding that `__all__` never mentioned.

## Behavior

Every commit is a move plus an import rewrite plus monkeypatch retargeting. Two adversarial passes checked that rather than assuming it:

- 1,407 functions AST-paired old-vs-new with module qualifiers normalised: 2 bodies differ, both the intended dependency sentinels. No changed defaults, narrowed `except` clauses, dropped `finally`s, or reordered side effects.
- 57 exception names resolved to runtime objects: none has split identity, so no handler catches a different class than the one raised.
- Decorators and `with`-contexts compared per function, plus `__wrapped__` across 847 functions: no decorator silently stopped wrapping.
- All 230 `monkeypatch.setattr` statements neutralized one at a time and diffed against the same sweep on `main`: no fault injection went dead. The two that resist neutralization were proved live by source perturbation instead.

Three dead monkeypatches *were* found and fixed: the plan assigned one to the wrong task, and four `test_routes.py` fault injections call `store.proposals.project` directly rather than through a route.

## One real bug, found and fixed

Hoisting `claude_agent_sdk` made `stream()` re-raise a *module-level* exception instead of constructing a fresh one per call. Python appends the raise site to `__traceback__` every time, so with the `claude` extra installed-but-broken the traceback grew `4 → 6 → 8 → 10 → 12` frames — and each retained frame pinned the scene transcript being sent. Walking `__traceback__.tb_frame.f_locals` recovered prior prompts verbatim. `stream()` now raises a fresh object; frame counts are flat and nothing is retained.

## Integrating with the new lock-domain guard

`test_lock_domain_guard.py` builds its `decorating` set from the *current file's* functions, so a decorator defined in a sibling module is invisible to it. This branch moves `@_serialized` into `scenes/locking.py` while its 18 users live in `scenes/write.py`, `moment.py` and `lifecycle.py`. They serialize — the decorator is unchanged — but the guard could not see it.

The guard is **extended, not exempted**: it now resolves a decorator reached through a sibling module and applies the same test it applies to a same-module one. The rejecting case is proved — with the campaign lock removed from `_serialized`, the guard fails naming all 17 mutators — and that probe is frozen as four tests (unlocked sibling, wrong-id binding, sibling rebinding `locks`, unresolvable receivers). Disabling the extension re-fails the positive test, so neither half is vacuous.

`UNREVIEWED` is re-keyed onto the new submodule names, not grown: the new mutator sets are subsets of the existing entries'. `# lock-domain-ok:` markers remain at exactly 2, both relocated onto the functions they exempt.

**Known coverage shrink, documented in the guard's docstring and not fixed:** `appearances.leave`, `set_base` and four `sheets` mutators left the guard's *survey* when the split moved their writes behind sibling helpers. That is lost coverage, never a false pass — all six were read and confirmed to still take the campaign lock. Following cross-module calls would be a much larger change.

## Verification

`check-lint`, `check-templates` (54 checks byte-for-byte), `check-web` (57 files / 681 tests) and `check-pydantic1` all pass. The suite reports `1 failed, 2925 passed`; the single failure is `test_atomic.py::test_a_read_only_record_is_not_silently_replaced`, which also fails on `main` in this container because `chmod 0444` does not stop uid 0 writing. Not a regression.

`check-pydantic1` matters here specifically: it runs the whole suite against the Android dependency set (pydantic 1.10, no `desktop` extra), which is what the `tiktoken` and `claude_agent_sdk` sentinels exist to keep working.

**The two CLAUDE.md Codex gates have not been run** — the Codex CLI is unavailable in this environment. Two Claude-driven adversarial passes were run in their place, but they are a substitute, not the gate. Please run `/codex:review` and `/codex:adversarial-review` before merging.


charlesmsiegel:
Thanks — all four P2s on `test_import_guard.py` check out factually. I verified each rather than taking them on faith, and all four are being carried into **Task 15**, which is the task that rewrites this exact file (it deletes the ratchet and makes the rule absolute). Batching them there avoids editing the guard twice while thirteen splits are still landing against it.

Per-finding assessment:

**L179 — trace re-exported functions before allowing by-value imports.** Agreed, real. `_kind_of()` walks only the target module's own top-level `FunctionDef`/`ClassDef`, so a function a leaf *re-exports* reads as `"other"` and escapes rule 3's by-value check — which is exactly the hazard the rule exists for. Will fix in Task 15.

**L188 — honor import markers when constructing cycle edges.** Agreed, real. Module-scope edges never enter `flagged`, so `# import-ok:` cannot exempt a cycle, yet the assertion message advertises it as the escape hatch for all three rule violations. It fails *closed*, so nothing unsound gets through — the defect is that the message misleads. Task 15 fixes the wording rather than adding a cycle escape hatch: a cycle is a structural problem and shouldn't have one.

**L258 — compare ratchet entries with multiplicity.** Real in principle, but it cannot bite here and stops existing shortly. I checked: the current 58-line baseline has zero duplicate strings (`sort | uniq -d` is empty), so no violation is currently masked by another. More to the point, Task 15 deletes the baseline and the membership checks entirely — the guard's final form asserts `_violations()` is *empty*, and multiplicity is meaningless against zero. Re-engineering the ratchet to use `Counter` would harden a mechanism with ~13 tasks left to live. Noting it and closing rather than fixing.

**L141 — restrict facade exemptions to local re-exports.** Correct on the facts — `store/climates/__init__.py:14` really does `from ..paths import home`, binding a function by value. Two things make me want a human call before changing it rather than just adopting it: the blanket `__init__.py` exemption is explicitly mandated by the design spec (`Enforcement` rule 4: "`__init__.py` files are exempt — re-exporting is their job"), and the concrete patchability harm doesn't land here, since backend tests isolate the store via `monkeypatch.setenv("GRIMOIRE_HOME", ...)` rather than by patching `paths.home`. Tightening the rule would surface pre-existing violations in modules this refactor doesn't otherwise touch, which is scope growth. Flagged for the repo owner to decide at Task 15.

Tracked in the branch's ledger so none of these get lost between here and Task 15.

---

charlesmsiegel:
Second round — one of these is being fixed now, two are parked with reasons.

**`store_api_baseline.json:2` — assert the captured baseline in the test suite. Correct, and being fixed now.** I verified it: `grep -rn store_api_baseline backend/ --include=*.py` returns nothing. The snapshot is committed but no test reads it, so the only comparison is a shell command pasted into each task's verification steps — meaning `pytest`, `make check` and CI all stay green if a split leaks or drops a facade attribute. Since "the public API is frozen" is a stated invariant of this refactor and there are twelve splits still to land against it, enforcement-by-human-discipline isn't good enough. A `backend/tests/test_store_api_baseline.py` is going in that compares both `__all__` and `dir()` and reports the symmetric difference per key, so a failure names the drifted symbol instead of dumping two hundred-element lists. Good catch — this was the difference between an invariant and an intention.

**`test_import_guard.py:92` — include parent-package initialization edges.** This is the sharpest finding so far and I'm not dismissing it, but I'm deferring the decision rather than patching it mid-refactor. You're right that `_resolve()` records only the leaf: `appearances.py` importing `campaigns.paths` executes `campaigns/__init__`, which imports `lifecycle`, which imports `appearances` — a real initialization cycle the file-granularity graph does not model. Our own independent reviewer of this same commit reached the identical conclusion unprompted, and traced the `store/__init__`-first, `overlay`-first, `scenes`-first and `sheets`-first import orders: each one currently resolves, because the facade binds `paths` and `read` before `lifecycle` and because `from .. import <partially-initialized>` falls back to `sys.modules`. So it works — but it works *because of* `campaigns/__init__.py`'s line order, and the design spec's explicit claim was that the submodule-binding rule makes `__init__` ordering stop being load-bearing. That claim is weaker than advertised. Two independent reviewers converging on it is strong signal, so it's recorded against the final review rather than closed. It's a spec-level question and the final graph shape isn't known until the `scenes` split lands; deciding it now would be guessing. The repo owner gets it with the traced evidence.

**`test_import_guard.py:236` — preserve cycle edges in the ratchet identity.** Real, and moot shortly. Encoding a cycle as its sorted SCC node set does mean a new edge inside the already-baselined `audit`/`scene_refs`/`scenes` component reproduces the same string. But that component is deleted from the baseline two tasks from now, when the `audit` split removes the cycle outright, and the final task deletes the baseline mechanism entirely — the guard's end state asserts `_violations()` is *empty*, where any cycle fails regardless of how its identity is encoded. Hardening a ratchet that has a handful of commits left to live would be churn. Noting and closing.

---

Issue #267, opened 2026-07-29
Two multi-campaign lock holders acquire in different orders (ABBA deadlock)


charlesmsiegel:
Surfaced while reviewing #255 (PR #265). Untouched by that change — it predates it.

`store/locks.py`'s ordering rules name the two actors that hold more than one campaign lock at a time:

> The module-edit lock is taken *before* campaign locks; multi-campaign holders (module publication, world-module rebind) are the only actors that hold more than one campaign lock […]

Those two acquire the set in **different orders**.

```python
# backend/src/grimoire/store/module_edit.py:504-511
@contextmanager
def _campaign_locks():
    """Every campaign's sheet lock: the User-edit vs LLM-play exclusion.
    Order is irrelevant — the module edit is the only multi-lock holder that
    can run concurrently with anything (LLM flows hold exactly one)."""
    with ExitStack() as stack:
        for c in campaigns.list_campaigns():          # sorted by `updated`, DESC
            stack.enter_context(locks.campaign_lock(c["id"]))
```

`campaigns.list_campaigns()` ends with `out.sort(key=lambda m: m["updated"], reverse=True)` (`store/campaigns.py:64`), so that is most-recently-updated first.

```python
# backend/src/grimoire/routes.py:1031-1036
all_cids = sorted(
    c["id"] for c in store.campaigns.list_campaigns() if c.get("world") == wid
)
with contextlib.ExitStack() as stack:
    for c in all_cids:                   # sole multi-lock holder; sorted order
        stack.enter_context(store.locks.campaign_lock(c))
```

That one is sorted by **id**.

## Why the module-edit lock doesn't save it

Every `_campaign_locks()` call site in `store/module_edit.py` (`:124` delete, `:259` journal replay, `:571` publish) runs inside `with _M:`. But `put_world_module` never takes `_M` — it goes straight to the campaign locks. So `_M` serializes module-edit operations *against each other* and does nothing to order them against the rebind route.

## The interleaving

Two campaigns in one world, `updated`-order and id-order disagreeing (the common case — ids are slugs, `updated` is recency):

| | thread A — `publish` / `delete_module` | thread B — `PUT /worlds/{wid}/module` |
|---|---|---|
| 1 | acquires `campaign_lock(recent)` | |
| 2 | | acquires `campaign_lock(alphabetically-first)` |
| 3 | blocks on `campaign_lock(alphabetically-first)` | |
| 4 | | blocks on `campaign_lock(recent)` |

[...]

Issue #234, opened 2026-07-28
Per-campaign locks are in-process only, but multi-device shared storage is advertised


charlesmsiegel:
Surfaced by the codebase atlas (`docs/codemap.html`, generated at 81b61ff1). **Ranked risk #2.**

## Problem

All campaign mutation serializes on a `threading.RLock` registry:

```python
# backend/src/grimoire/store/sheets.py:618-627
def lock_for(cid: str) -> threading.RLock:
    with _registry_guard:
        return _campaign_locks.setdefault(cid, threading.RLock())
```

`threading.RLock` protects nothing across processes. But the product actively encourages exactly that scenario — from the README:

> point the data directory at a synced folder (Dropbox, iCloud, etc.) to share one library across devices

The same store is also reachable from the desktop app and the Android APK simultaneously.

## Consequence

Two devices — or two backend processes on one machine — editing one campaign have **no mutual exclusion at all**. Because most records are rewritten whole rather than merged (see #233), the loser's edit is silently overwritten. There is no detection, no conflict marker, and no user-visible symptom.

The threat model is stated explicitly in `store/module_edit.py:4-10`:

> Concurrency threat model (spec): exactly two actors — the User (UI) and the LLM (play flows).

Both actors live inside one process. The sync-folder workflow adds a third the design does not account for.

## Options

1. **File-based locking** (`portalocker` / an atomic lockfile protocol) for the campaign lock domain, so the guarantee survives process boundaries.
2. **Optimistic concurrency**: the store already computes content hashes (`store/entities.py:130`) and the sync engine already does three-way comparison (`store/sync.py`). A compare-and-swap on write would detect the clobber even if it can't prevent it.
3. **Scope the claim down**: if cross-device concurrent editing is out of scope, say so in the README and detect-and-warn on a stale read rather than silently losing data.

Option 2 reuses machinery that already exists and degrades honestly; worth considering first.

PR #297, opened 2026-08-06
Hold the campaign-lock ordering rule down in code, not in prose (#267)

charlesmsiegel:
Refs #267. Left as a reference rather than `Fixes` — see "About closing #267" below; it's a one-word change if you'd rather it auto-close.

> **Scope note.** This PR carried an AST guard through nine review rounds; it was **removed** in `08953cf` at the maintainer's call. What ships is two files, +91/−31. The history is in "What the guard cost, and why it went" at the bottom — worth reading before anyone proposes rebuilding it.

## What #267 asked for, and what was actually left

The **runtime deadlock is already fixed on `main`**. `locks.hold_all` exists, sorts, and both multi-campaign holders go through it — it landed with the #234 cross-process lock work without referencing #267. So the issue is stale on its headline claim.

What survived is the part the issue itself calls the real defect: *"The comments are the real defect … Whoever touches either next reads a comment saying ordering can't matter."*

## 1. Three comments said the opposite of the truth

All three claimed the two multi-campaign holders had always agreed on acquisition order, "by accident", because `list_campaigns()` walks `sorted(base.iterdir())`. It does — and then ends on `out.sort(key=lambda m: m["updated"], reverse=True)`:

```
list_campaigns() order: ['mike', 'alpha', 'zulu']
sorted(ids)           : ['alpha', 'mike', 'zulu']     id-sorted? False
```

Recency, not id, while the rebind route sorts by cid. **The inversion was live**, not hypothetical, and `hold_all` is what closes it rather than a guarantee bolted onto an agreement that never existed. Corrected in `store/locks.py`'s module docstring, `hold_all`'s docstring, and the note above the order tests.

`locks.py` now also states what is *not* guaranteed: the rule is checked per holder, a third holder is checked by nothing, and adding one means adding its test. That is the honest version, and it replaces a sentence that claimed a structural guarantee.

## 2. The same mistake had reached the tests

Both order tests created their three campaigns inside one `updated` second. A stable sort on equal keys leaves them in `iterdir()` order — which *is* cid order — so the acquisition order came from the fixture rather than from the code under test.

Verified rather than theorised: with `module_edit._campaign_locks` reverted to its pre-fix raw `ExitStack` loop, `test_module_edit_acquires_campaign_locks_in_sorted_order` **passed**. Only the source-grep test failed.

Both now stamp `updated` through the store's own `touch`, so recency order is the reverse of id order, and they assert that precondition — if `list_campaigns` ever returns cid order again, the fixture says so instead of quietly going vacuous.

**Measured effect:** reverting `module_edit` to its pre-fix loop now fails **2** tests where it failed **1** before.

## Verification

`3590 passed`, ruff clean, `verify_templates.py` byte-for-byte.

One failure, pre-existing and environmental: `test_atomic.py::test_a_read_only_record_is_not_silently_replaced` fails identically on unmodified `main` in this container, which runs as uid 0 — root bypasses the read-only bit.

**Not run:** `check-web` (frontend untouched) and `check-pydantic1` (no pydantic in the diff). **GitHub Actions has never triggered a run on this PR** — the workflow's last run repo-wide was 2026-08-02, so this looks like an Actions-side pause rather than anything about this branch. Every number above is local.

## What the guard cost, and why it went

#267 also suggests a durable guard — "a source assertion … both multi-lock holders sort by id". That was built, generalised to *"only `hold_all` may hold more than one campaign lock"*, and carried through nine review rounds and **twenty findings**. Every finding was against the guard. None was against the `store/locks.py` change, which is untouched since the first commit.

The last round is what settled it: **three of its five findings were the guard rejecting *safe* code** — a lazily-created async generator expression, a comprehension-local target reusing the campaign's name, and a nested `def` whose body doesn't run where it's written. Confirmed by running the deleted guard against all three. It had no `# …-ok:` marker by design, so a false positive is a wall, not an annoyance.

Two things worth keeping from the exercise, for whoever revisits this:

- **The enumerating shape is a trap.** Every one of the twenty findings was the same hold in punctuation the detector hadn't been taught. `test_lock_domain_guard.py` documents this exact lesson — *"seven review rounds produced roughly eight findings of a single shape… So the polarity is inverted"* — and *"**A name is not a binding.**"* Three separate fixes here turned out to be lessons already written in that file.
- **The gap is real and now stated rather than enforced.** A third multi-campaign holder written tomorrow is checked by nothing. `locks.py` says so in as many words, which is strictly better than the comments that were there before — each claiming to be the sole multi-lock holder.

If a guard is rebuilt later, the two things worth reusing are the inverted polarity (whitelist the one permitted shape; everything else fails) and `test_lock_domain_guard.py`'s existing machinery, rather than a second implementation of it.

## About closing #267

Left as `Refs` rather than `Fixes`. The comments and the vacuous tests are fixed; the durable guard the issue also suggests is deliberately **not** shipped, so #267 may be worth keeping open with that narrowed scope. Your call.


charlesmsiegel:
All four Codex findings were real. Confirmed each against the detector before changing anything, then fixed in `36b2c5c`.

| # | Finding | Verdict |
|---|---|---|
| 1 | multi-item `with` through locals | miss — confirmed |
| 2 | unrolled `acquire()` pair | miss — confirmed |
| 3 | comprehension over `enter_context` | miss — confirmed |
| 4 | sequential `ExitStack` blocks | **false alarm** — confirmed |

Finding 4 is the one I'd have most regretted shipping: the guard has no `# …-ok:` marker by design, so rejecting deadlock-free code is a wall rather than an annoyance.

**The fix is not four patches.** The pattern across these four — and the two holes the pre-push review found — is that the detector was a *list of bad spellings*, and the list was always one spelling short. It now keys on **what releases a hold**:

- stack registrations are grouped by the `with` block bounding their stack, so two in one block collide and two sequential blocks don't (4);
- bare `acquire()`s are grouped as bounded by nothing, so any two in a scope collide (2);
- loops *and comprehensions* both mark a hold as repeating (3);
- the multi-item count uses `_is_lock` rather than `_lock_calls`, so a bare `Name` bound to a lock counts (1);
- any hold inside a lock-`with` body counts against it — which caught a sixth shape none of the findings named, a stacked lock nested inside a `with`.

All six are now permanent self-tests, with the review that found them cited in the comments. Mutation matrix unchanged: `module_edit` reverted to its pre-fix loop fails 3 tests, a new third holder fails 1, clean tree fails 0. Full suite `3617 passed`, ruff clean.

The one remaining failure, `test_atomic.py::test_a_read_only_record_is_not_silently_replaced`, is pre-existing and environmental — it fails identically on unmodified `main` in this container, which runs as uid 0, and root bypasses the read-only bit.

Two limits I chose not to close, both now recorded in the guard's reach list rather than left implicit: branches and statement order aren't tracked (two `acquire()`s count as two even with a `release()` between them), and two *nested* `ExitStack`s group separately though both are live. Closing either is dataflow analysis; neither shape appears in the tree.

---

charlesmsiegel:
All three findings were real — confirmed against the detector before changing anything:

| Finding | Verdict |
|---|---|
| annotated lock locals (`AnnAssign`) | miss — confirmed |
| registration live across a sibling `with` | miss — confirmed |
| same stack registered under an unrelated `with` | miss — confirmed |

**But I did not fix them.** That would have been the eighth spelling in four rounds, and the pattern is the finding:

```
lock = campaign_lock(c) then stack.enter_context(lock)   (a local)
with campaign_lock(a), campaign_lock(b):                 (one statement)
first: Lock = campaign_lock(a) ...                       (AnnAssign)
campaign_lock(a).acquire(); campaign_lock(b).acquire()   (unrolled)
[stack.enter_context(campaign_lock(c)) for c in cids]    (comprehension)
a registration still live across a sibling `with`        (across forms)
the same stack registered under an unrelated `with`      (grouping)
```

Seven findings, one defect: the detector asked *"does this look like a multi-hold I recognize?"* and the list of recognized spellings was always one short. `test_lock_domain_guard.py` in this repo hit the same wall — "seven review rounds produced roughly eight findings of a single shape" — and solved it by **inverting the polarity**. So this guard now does the same (`c0c9dd0`).

**The rule.** Outside `hold_all`, in any one scope: every `campaign_lock(...)` / `hold_all(...)` call must be the direct context expression of a `with`, and they must all name the same campaign. Nothing else is accepted.

`AnnAssign`, `.acquire()`, comprehensions and stack registrations need no special case — none is a `with`, so none passes, and neither will the eighth spelling. **~20 lines where the enumerating version was ~110**: no locals resolution, no parent map, no lifetime grouping, no accumulator list. All thirteen shapes from all four rounds are now self-tests, and the diff is net **−139 lines**.

**Two safe shapes are now rejected**, stated in the comment rather than left to be discovered:

- a lock bound to a local before its `with`;
- two sequential `with ExitStack()` blocks holding one lock each — your finding 4 from the previous round. It is still rejected, but now *by the stated rule* rather than by an accident of grouping, and it is a self-test rather than a claim.

Both rewrite to a `with` in one line, which is why there is still no `# …-ok:` marker.

**Deliberately allowed:** two `with campaign_lock(cid)` blocks naming the same campaign in one scope. `routes/streaming.py` does this in `finalize`'s mutually exclusive branches, and a campaign's lock cannot deadlock against itself. A stricter draft flagged it; that was a false alarm on real code and is now a self-test too.

Mutation matrix unchanged — pre-fix loop 3 failures, new third holder 1, clean 0. Full suite `3612 passed`, ruff clean. The single remaining failure is the pre-existing uid-0/read-only-bit one that also fails on unmodified `main`.

---

charlesmsiegel:
Both findings were real — confirmed against the detector before changing anything. Fixed in `743759d`, and all nine review threads on this PR are now resolved.

These two are a **different kind of gap** from the seven before them. Those were ways to *hold* a lock the detector didn't recognize, and inverting the polarity closed them. These are ways to *name* one — which is the other lesson `test_lock_domain_guard.py` had to learn and write down:

> **A name is not a binding.** It was fixed four separate times before it was written down.

So the same inversion is applied to identification. The whitelist is now over **mentions** of the protected names — `Name`, `Attribute`, or string constant — and the only permitted mention is the callee of a `with` statement's context expression.

- A rename is an **offender**, not something to resolve, so there is no alias to chase. `cl = locks.campaign_lock` and `from …locks import campaign_lock as cl` both fail; importing under the factory's own name still reads as itself.
- `getattr(locks, "campaign_lock")` falls out for free — nobody named that one, and it is now a self-test.
- `_campaign_of` reads the `cid`/`cids` keyword as well as the positional, and a call whose campaign can't be read gets a **fresh unknown that compares equal to nothing**. Unknown means fail. Reading two unreadable calls as one campaign is exactly what let the keyword pair through.

Seventeen shapes are self-tests now. Mutation matrix unchanged — pre-fix loop 3 failures, new third holder 1, clean 0. Full suite `3618 passed`, ruff clean; the single remaining failure is the pre-existing uid-0/read-only-bit one that also fails on unmodified `main`.

**Running total: eleven findings across five rounds, every one against the guard rather than the `store/locks.py` change.** Both times the fix turned out to be a lesson this repo had already written down and I was re-deriving. Worth saying plainly, since the guard's value is that the next person doesn't repeat it either.

---

charlesmsiegel:
Real — confirmed before changing anything. Fixed in `939d85e`; thread resolved, so all ten on this PR are now closed.

This is the naming half again, one level down: not the name of the *factory* but the name of the *campaign*. `ast.dump(Name('cid'))` twice is one campaign only if `cid` still means what it meant the first time.

Third finding in a row whose answer was already written in this repo — `test_lock_domain_guard.py` needs the identical fact and calls it `_rebinds_cid`:

> every rule here reads that name as *this function's campaign*, which is a claim about the parameter

The claim is false when the scope rebinds it. So `_campaign_of` now returns a fresh sentinel for a campaign expression built from any name the scope assigns to — `Assign`, `AnnAssign`, `AugAssign`, walrus, `for` target, `with … as` — alongside the sentinel it already returned for an unreadable argument. Unknown compares equal to nothing, so two unknowns are two campaigns. Same polarity that closed the keyword pair.

**The boundary matters and is a test.** A scope holding *one* campaign lock is unaffected, since one campaign cannot disagree with itself — which is exactly what keeps the sequential `for c in ids: with campaign_lock(c):` legal, though its target is rebound every iteration. Both that and "one lock whose name is rebound elsewhere" are in the no-false-alarm set.

Nineteen shapes are self-tests. Mutation matrix unchanged — pre-fix loop 3 failures, new third holder 1, clean 0. Full suite `3620 passed`, ruff clean.

**Twelve findings across six rounds, all against the guard, none against the `store/locks.py` change.** Three of the last four had their answer sitting in `test_lock_domain_guard.py`, which is either an argument that these two guards should share machinery, or an argument that its comments are the best documentation in the repo and I should have read them first. Probably both — but folding them together is a bigger change than #267 should carry, so I've cited it in place instead.

---

charlesmsiegel:
Both real — confirmed before changing anything. Fixed in `94fe9f0`; both threads resolved, so all twelve on this PR are closed.

**Stateful expression.** `with campaign_lock(cids.pop()):` nested inside itself: identical dumps, different campaigns, and no name assigned for the rebound check to notice. `_campaign_of` now **pins** a campaign only for a constant or a name the scope never rebinds; everything else is unknown, and unknown never equals unknown. That *subsumes* last round's rebound check — a rebound name simply isn't pinned — and closes `d[k]`, `next(it)`, `x.y` and whatever the next one would have been. The enumerating instinct showed up one last time in `_campaign_of`, and this is the same inversion applied to it.

**Suspension.** A generator paused inside `with campaign_lock(cid):` still holds it, so two live instances hold two campaigns in whatever order the consumer advanced them — one syntactic acquisition, two locks. That's now a third clause on the rule: no `yield` or `await` while a campaign lock is held.

The exception is a `@contextmanager`, whose yield *is* its caller's `with` body. That isn't a carve-out invented to make the suite pass — it's the only shape the package uses, and both instances are exactly it:

- `proposals.locked` (`store/proposals.py:88`)
- `module_edit._campaign_locks` (`store/module_edit/migrate.py:311`)

I checked every suspension inside a campaign-lock `with` in the package before writing the rule; those two are the complete list. Both spellings of the decorator (`@contextmanager`, `@contextlib.contextmanager`) are self-tests.

`await` wasn't reported. It's the same extended lifetime, nothing in the tree does it, so rejecting it now costs nothing — also a self-test.

Twenty-five shapes are self-tests. Mutation matrix unchanged — pre-fix loop 3 failures, new third holder 1, clean 0. Full suite `3626 passed`, ruff clean.

---

**Fourteen findings, seven rounds, all against the guard.** Worth being straight about the trend rather than just reporting green: rounds 5–7 each found something whose fix was *the same inversion applied one level deeper* — first to how a lock is held, then to how the factory is named, then to how the campaign is named, and now to when the hold ends. That is a converging sequence, not an endless one, and each step deleted special cases rather than adding them. But I'd rather say plainly that I can't prove it's finished than claim it is. The reach limits are written into the guard's header, and the `store/locks.py` change this PR exists for has been untouched since the first commit.

---

charlesmsiegel:
All four real — confirmed before changing anything. Fixed in `a3df0ce`; threads resolved, so all sixteen on this PR are closed.

Three are the same mistake in three places — something the guard treated as fixed that isn't:

- **`hold_all(cids)`, `cids.append(low)`, `hold_all(cids)`** — the binding holds still while the contents move. A *set* of campaigns is now never pinned. A single `hold_all` is unaffected: it sorts its own argument, and one key cannot disagree with itself.
- **`with hold_all(cid), campaign_lock(cid):`** — the sharpest of the sixteen. Identical arguments, and `hold_all` iterates a string into one lock *per character* while `campaign_lock` takes the whole id. The campaign key now includes the factory.
- **`with campaign_lock(cid), await make():`** — a `with` enters items left to right, so the await suspends with the lock held. The scan covered only the body; it now covers every item after the lock-bearing one. Items *before* it hold nothing yet, and that boundary is a test.

**The fourth is a genuine over-exemption I introduced last round**, and it's the one I'd have least liked to keep. Treating `@contextmanager` as bounded suppressed *every* suspension in the function, not just the yield it was meant to permit. An `@asynccontextmanager` awaiting during setup or teardown hands control to the event loop with the lock still held — and the `RLock` half is owned by a **thread, not a task**. Only a `yield`, and only in a context manager, is exempt now; an `await` never is.

That one is worth flagging on its own: the previous round's fix created it. Six rounds of findings had been pre-existing gaps; this is the first time a fix of mine opened a new hole, which is exactly what the exemption's breadth should have made me check at the time.

Thirty-one shapes are self-tests, including both boundaries these fixes turn on. Mutation matrix unchanged — pre-fix loop 3 failures, new third holder 1, clean 0. Full suite `3632 passed`, ruff clean.

---

**Eighteen findings, eight rounds, all against the guard.** I told the user last round the sequence was converging. That was too confident: three of these four were new *kinds* of unpinnable thing rather than deeper applications of the same idea, and one was self-inflicted. The `store/locks.py` change this PR exists for is still untouched since the first commit, and remains the part I'd actually stake a claim on.

If you'd rather stop paying for this, there is a smaller version available: keep the three corrected comments and the two de-vacuumed order tests, and drop the AST guard entirely. That is #267's literal ask minus the durable part, it is ~40 lines instead of ~250, and it has needed no review rounds at all. Happy to cut it that way if the guard is costing more than it's worth to you.

---

charlesmsiegel:
Both real — confirmed before changing anything. Fixed in `1ad925a`; threads resolved, so all eighteen on this PR are closed.

These two are the same recurring failure — an enumeration that came up one short — but with an important difference from the previous sixteen: **they enumerate over Python's grammar, which is finite and knowable.** So they can be *closed* rather than extended.

**`_rebound` no longer lists statement types.** It asks whether the AST calls a `Name` a store. That covers assignment, augmented and annotated assignment, `for`, `with … as`, comprehensions, the walrus, match captures — and anything a future Python adds in that form. What remains is the short list of binders carrying a bare `str` rather than a `Name` node: `except … as`, `case … as`, `case [*rest]`, `case {**rest}`, import aliases, `global`/`nonlocal`, and a nested `def`'s own name. That list is closed because the grammar says so, not because I stopped thinking of more.

**`_SUSPENDS` gains `AsyncFor` and `AsyncWith`**, plus async comprehensions — matched on the comprehension *expression*, since the `ast.comprehension` clause that carries `is_async` has no `lineno` to report (found by the first version raising `AttributeError`, not by reasoning). With `Await`, `Yield` and `YieldFrom` that is every suspension point the language has.

The synchronous twins are now self-tests too — `for`, `with`, a plain comprehension suspend nothing and must stay legal inside every critical section in the store. Flagging those would have been a far worse outcome than the miss.

Thirty-eight shapes are self-tests. Mutation matrix unchanged — pre-fix loop 3 failures, new third holder 1, clean 0. Full suite `3639 passed`, ruff clean.

---

**Twenty findings, nine rounds.** I want to be straight about the shape of this rather than just reporting green again, since I've now twice told the user it was converging and been wrong.

What is true: every finding has been against the guard, never against the `store/locks.py` change this PR exists for, which is untouched since the first commit. And the last two were closable in a way the earlier ones weren't — the grammar is finite in a way "ways a programmer might spell a lock hold" is not.

What I can't claim: that this is the last round. The offer from my previous comment stands — drop the AST guard, keep the three corrected comments and the two de-vacuumed order tests, and you have #267's literal ask in ~40 lines that has attracted zero findings. I'd still keep the guard, but that's a preference, not evidence, and it's your review budget being spent.

---

charlesmsiegel:
**The AST guard is removed** (`08953cf`), at the maintainer's call. All 23 review threads are resolved.

The five findings on `1ad925a` arrived after that decision and are moot — but I checked them against the deleted guard rather than waving them off, because three of them are the strongest evidence that removing it was right:

| Finding | Verdict |
|---|---|
| lazy async genexp reported as a suspension | **rejected safe code** |
| comprehension-local target reusing the campaign name | **rejected safe code** |
| nested `def` body scanned as if it ran there | **rejected safe code** |
| `@asynccontextmanager` yield exempted | miss |
| `backend/scripts/` outside the scan | coverage gap |

All three false alarms confirmed by running the deleted guard against them. With no `# …-ok:` marker by design, a false positive is a wall rather than an annoyance — the guard had started forbidding valid code, which is a worse failure than the class of bug it caught.

## What ships

Two files, +91/−31:

- **The three comments corrected.** They claimed the two multi-campaign holders had always agreed on acquisition order "by accident". They hadn't: `list_campaigns()` walks `sorted(base.iterdir())` and then re-sorts by `updated`, so it returns recency while the rebind route sorts by cid. That was the defect #267 named.
- **The two order tests made able to fail.** Both created their campaigns inside one `updated` second, so a stable sort left them in iterdir order — which *is* cid order — and the assertion was satisfied by the fixture. They now stamp `updated` through the store's own `touch` and assert that precondition. Measured: reverting `module_edit` to its pre-fix loop now fails **2** tests where it failed **1**.
- **`locks.py` no longer claims a guarantee it doesn't have.** It says the rule is checked per holder, that a third holder is checked by nothing, and that adding one means adding its test.

`3590 passed`, ruff clean, templates byte-for-byte.

## For whoever revisits this

The guard drew twenty findings across nine rounds; the `store/locks.py` change drew none. If it's rebuilt, two things are worth carrying over rather than rediscovering:

1. **Invert the polarity** — whitelist the one permitted shape, fail everything else. Enumerating bad spellings was wrong twenty times out of twenty.
2. **Reuse `test_lock_domain_guard.py`'s machinery.** Three of the fixes here turned out to be lessons already written in that file — *"seven review rounds produced roughly eight findings of a single shape"* and *"**A name is not a binding.**"* A second implementation of it is what this PR tried and abandoned.

---

PR #310, opened 2026-08-06
First-run setup wizard (#194)

### LINKED ISSUE #194: First-run setup wizard
Part of the **Frontend Shell, HUD & Views** milestone. From the pre-rebuild target feature list:

> First-run setup wizard

## Current state
There is no first-run detection or wizard today — a fresh install lands on `CampaignsView` (`/`) with no worlds and no campaigns, and the user must discover on their own that they need to visit `/worlds` to create a world and `/config` to set an OpenRouter key. No onboarding/welcome component exists anywhere in `frontend/src` (verified: no "onboard"/"welcome"/"first run" hits).

The pieces a wizard needs to stitch together already exist independently:
- **API key gating**: every LLM route calls `_require_key(cfg)` (`backend/src/grimoire/routes.py:1303`), which raises `409 {"detail": "OpenRouter key not set", "kind": "missing_key"}` when `cfg["openrouter_key"]` is empty. The frontend doesn't catch this `kind` specifically anywhere — instead `App.tsx` fetches `GET /config` once at boot into a `keySet` boolean, threads it down as a prop, and `CampaignView.tsx:411` renders a banner ("No OpenRouter key set. Set your key in Config.") whenever `!keySet`; `CampaignWizard.tsx:337,341` similarly disables its "Generate" opener button and shows a hint when `!keySet`. So the key requirement is surfaced reactively per-view, not as a blocking first-run gate.
- **Storage location**: `GET/PUT /config/data-dir` (`routes.py:278-289`, backed by `store.data_dir_info()`/`store.set_data_dir()`) is already a full feature in `ConfigView.tsx` ("Storage location" section: shows the current dir, lets you move it, disabled when set via `GRIMOIRE_HOME` env, has a "Reset to default" link).
- **Theme**: `ConfigView.tsx`'s "Theme" section already lists `themeList` cards and calls `save({theme})`, which round-trips through `PUT /config` and `ThemeProvider`.
- **First world/campaign**: `WorldsView.tsx` (create-world form) and `CampaignWizard.tsx` (`/campaigns/new`, 4-step: Backdrop → Character → Locations → Opening, `STEPS` array, `wizard-steps`/`wizard-body`/`wizard-footer` CSS classes, a `commit()` that creates the campaign, seats a PC, and offers greeting/opener generation) are the existing patterns to reuse, not rebuild.

### Option A — first-run wizard as an orchestrator over existing views (recommended)
- Add first-run detection in `App.tsx`'s boot `useEffect` (already calls `api.getConfig()`): treat "no key set AND zero worlds" (an extra `api.listWorlds()` check) as first-run, and route to a new `/welcome` wizard instead of `CampaignsView`.
- Build the wizard as a 4-step flow mirroring `CampaignWizard`'s step-array pattern: (1) OpenRouter key — reuse the key `<input type="password">` + `PUT /config {openrouter_key}` from `ConfigView`; (2) storage location — reuse the "Storage location" block verbatim (it's small); (3) theme — reuse the `theme-cards` block; (4) hand off into the existing `WorldsView` create-world form, then straight into `CampaignWizard` for the first campaign.
- Tradeoff: most reuse, least new UI code, but requires factoring the key/storage/theme blocks out of `ConfigView.tsx` into shared sub-components so both `ConfigView` and the wizard render the same markup instead of duplicating it.

### Option B — standalone wizard component, `ConfigView` untouched
- Write a new `SetupWizard.tsx` that duplicates (rather than extracts) the key/storage/theme form fields inline, then calls the same `api.putConfig`/`api.putDataDir` methods `ConfigView` uses.
- Faster to ship (no refactor of `ConfigView`), but the two forms will drift over time since they're not sharing markup or validation.

## Implementation notes
- Detecting "first run" from `key_set === false` alone is not sufficient — a returning user who cleared their key shouldn't be re-forced through a full wizard every load; pair it with an empty-worlds check (or a one-time `localStorage`/config flag) so the wizard is skippable/dismissable.
- `_require_key`'s `409 {kind: "missing_key"}` is currently never pattern-matched in the frontend (`ApiError` just carries `.detail`) — if the wizard is meant to also catch this reactively (e.g. a user skips setup then tries to chat), that `kind` field needs a first frontend consumer.
- Frontend tests: from `frontend/`, `npx vitest run` + `npx tsc -b`. Cover: first-run detection routes to the wizard when no key + no worlds; the wizard's key/storage/theme steps call the same `api.putConfig`/`api.putDataDir` methods `ConfigView.test.tsx` already exercises; finishing the wizard lands in `CampaignWizard` or `CampaignView`.

## Dependencies & related
- #191 — a persistent nav sidebar changes where "first run" would even land (e.g. a Library hub instead of `CampaignsView`).
- #39 — "Guided world hub" (setup checklist, per-kind counts, suggested next actions) is a related but distinct ongoing-guidance feature, not a one-time wizard; the two should share language/design but not be conflated.

charlesmsiegel:
Fixes #194.

A fresh install landed on an empty campaigns list and left the user to work out on their own that a world had to be made in `/worlds` and a model connected in `/connections`. It now opens a four-step wizard at `/welcome` instead.

## The theory

**An install is "fresh" when its store has never recorded an answer to setup *and* holds no worlds or campaigns.** Both halves are load-bearing: the flag alone would show the wizard to every existing install on upgrade, and emptiness alone would re-trap someone who deleted everything, or nag someone who skipped.

The two ways to be wrong are not symmetric — hijacking an existing library is far worse than missing a fresh install — so every uncertain case (unreadable store, failed write) answers `first_run: false`.

## Backend

- New `setup_done` config key and a derived `first_run` on `GET`/`PUT /api/config`, so the boot fetch the frontend already makes answers the question in one request rather than three, and an unreadable store fails closed instead of looking empty.
- `worlds.has_worlds()` / `campaigns.has_campaigns()` over `paths.any_child_record()` — the same filter the listings apply, but stopping at the first hit and parsing nothing. It probes with `stat` rather than `is_dir()`/`exists()`, which answer `False` for a permission error exactly as they do for absence; a library that merely could not be opened must not read as empty.
- `setup_done` is recorded when the first world or campaign is **created**, not only when a config read notices content. The backfill on read still exists for stores that predate the key.
- `store.config.write_config` takes a new `locks.config_lock()`. It is a read-merge-write of one file, and this branch puts a writer on the read path (the backfill), so without serialization a second tab merely *loading the app* can erase a setting saved in the first.
- The config response carries the `data_dir` it describes, so a client can tell that a cached decision belongs to a library it is no longer looking at.

## Frontend

`/welcome` renders `SetupWizard`; **only `/` redirects**, so deep links stay honoured and the topbar remains an escape hatch. Both routes branch on one `inSetup`, so they are exact complements and cannot bounce a redirect between them.

Steps are ordered by dependency: **storage** (everything after is written into it) → **connection** → **theme** → **first world** (the handoff into `CampaignWizard`, whose own first step needs a world). Each step commits as it is answered, so abandoning the wizard never discards work the user watched succeed — the trade-off being that Back is navigation, not undo.

`firstRun` is the server's live verdict, refreshed on every navigation with a cache-bypassing read. A separate latch records that setup was *left*, keyed to the store it was left in: `finish()` writes `setup_done` best-effort so a failure cannot strand anyone, which means the verdict alone could redirect straight back into the wizard forever.

Every control holds while any of the wizard's four writes is in flight — a config save that overlaps another makes the outcome depend on response order, and navigating away from one unmounts the only place its failure would be reported.

## Reuse, not duplication

Per the issue's Option A, `StorageLocation`, `ThemePicker` and `ConnectionForm` are extracted out of `ConfigView` / `ConnectionEditor` rather than re-typed, so the two places cannot drift.

Also fixes `button.link`, which was scoped to `.config` and so rendered as a default chrome button everywhere else it is used — the Connections page's "Fetch models" link already had this.

## Deviations from the issue, and why

The issue text had drifted from the code since it was filed:

1. **"OpenRouter key" step is a "Connection" step.** The cited `routes.py:1303` and `cfg["openrouter_key"]` no longer exist — the single key was replaced by `llm-connections` (three kinds), and `_require_key` moved to `routes/common.py`.
2. **Storage comes first, not second.** Everything the later steps write lands in whichever store step 1 selects.
3. **Detection moved server-side.** The issue proposed a client-side "no key AND zero worlds" with an extra `api.listWorlds()`; `first_run` replaces it.
4. **The world-create form is not extracted.** It is one labelled input and a button; sharing it would be an abstraction over a single field.

Deliberately **not** built: a first frontend consumer for `409 {kind: "missing_key"}`. The issue frames it conditionally, and a one-time gate does not need reactive catching.

Known and accepted (owner's call): when `GRIMOIRE_HOME` is set — every Android launch — the storage step's control is disabled and its hint says to unset the variable. The step still shows where the library lives, and that hint is the same copy the Config page already shows.

## Review

**Codex reviewed this five times and raised 24 findings. All 24 are fixed.** An earlier version of this description said the Codex gates could not run because the CLI was not installed in the authoring container; that was wrong — Codex is wired into this repo as a PR reviewer and the gate ran on every push. The note is removed rather than left to mislead.

Three of those findings were corrections to this branch's own earlier fixes, and two were reversals of positions taken in the thread:

- The `write_config` lock was deferred for three rounds as "pre-existing". It was not: this branch added a writer on a `GET`, which is what makes the lost update reachable without the user doing two things at once.
- "Refresh first-run state" was declined twice before being implemented. The objection (a live verdict traps a user whose flag write fails) was only ever an argument against refreshing *without a latch*, not against refreshing.

Also caught by review, not by the author: `is_dir()`/`exists()` swallowing permission errors, which made an unreadable library read as empty and open the wizard over someone's real collection — the exact failure direction this PR's own docstring declared impossible.

## Verification

| Gate | Result |
|---|---|
| `check-py` | 3860 passed, 1 failed |
| `check-web` | 60 files, 898 tests passed |
| `check-lint` | clean |
| `check-templates` | 62 checks passed |
| `check-pydantic1` | run on an earlier commit: 3605 passed, same 1 failure |

The one failure is `test_atomic.py::test_a_read_only_record_is_not_silently_replaced`: it chmods a file to `0444` and expects `PermissionError`, but the authoring container runs as uid 0 and root writes through permission bits. It reproduces identically on unmodified `origin/main` in a scratch worktree, and this branch does not touch `store/atomic.py`.

Every fix carries a test, and each was mutation-checked by reverting the fix and confirming its test fails. That check earned its keep twice: in one round three tests did not actually pin their fix (one passing only because a default mock made the wrong expression coincidentally right), and in another the concurrency test passed with the lock removed, because `TestClient` serializes requests across threads and the "concurrent" interleaving never happened.

Driven end-to-end in Chromium against an isolated store: `/` → `/welcome`, connection created and activated (topbar flipped to `CONNECTED`), theme persisted, world created, handoff into `/campaigns/new` with the world present, `setup_done: on` on disk, and a reload at `/` stayed put.

`origin/main` is merged in as of 3070061 — it had moved 36 commits ahead across five files this branch also touches, `store/locks.py` most consequentially.


charlesmsiegel:
Four of the five findings were real and are fixed in b3da58f. One is declined, with reasoning.

**Fixed**

- **Gate the welcome route** — reachable, and the duplicate-world path is the sharp end of it: reload after step 4 has created a world and the store is no longer a first run, but the route mounted the wizard anyway and restarted at step one. Both routes now branch on the same `firstRun`; being exact complements is what keeps them from bouncing a redirect between them.
- **Validate connection readiness** — correct, and the message was the lie: a name alone saved and activated a keyless OpenRouter connection, then printed "Connected ✓" over something `_connection_ready` reports as `ready: false`. The gate now mirrors that function kind for kind (key for `openrouter`, base URL for `openai_compatible`, neither for `claude`).
- **Wait for the storage change** — the hidden-failure half is the one that bites: unmounting `StorageLocation` mid-request discards the only place the error would have appeared. It now reports pending through a new `onPending`, and the step holds Next until the move settles.
- **Finish saving the theme** — confirmed. `write_config` reads the raw frontmatter, merges and rewrites, with no lock, and FastAPI runs these sync endpoints in a threadpool, so two overlapping PUTs genuinely lose one. Next, Back and Skip are all held while the theme save is in flight.

**Declined — refresh first-run state after changing stores**

The staleness is real, but refreshing the verdict on every navigation trades it for a trap. `finish()` treats the `setup_done` write as best-effort so a failure can't strand anyone on the wizard; with a live verdict, a store that cannot record the flag answers `first_run: true` on the next refetch, `/` redirects back into the wizard, and skipping repeats that forever. The boot-time verdict guarantees one skip always gets the user out.

What that costs is bounded: an established install that repoints Config at a fresh store sees the campaigns list rather than the wizard until the next launch, and a fresh install pointed at an initialized store is asked to create a world it doesn't need and can skip. Neither loses data or blocks anything. Happy to revisit if you'd rather have the live verdict and a separate "dismissed this session" latch to keep the escape hatch.

**Adjacent, not fixed here**

The lost update behind the fourth finding is not specific to this branch — `store.config.write_config` takes no lock, so *any* two overlapping `PUT /api/config` calls can drop one. `ConfigView`'s quote-color checkbox races its own theme click on `main` today. Serializing config writes belongs in its own change rather than as scope creep here; say the word and I'll open the issue.

Each fix carries a test, and I mutation-checked them: reverting the four guards fails exactly the five new tests and nothing else. Full frontend suite is 60 files / 863 tests green.

---

charlesmsiegel:
All four of the second round are real and fixed in 740af4e. One of them corrects a rebuttal I made in the first round.

- **Serialize theme selections** — right, and my previous fix was half a fix: disabling Next while `savingTheme` was true left the cards themselves clickable, so two picks still start two concurrent writes and the first response clears the single flag while the second is in flight. `ThemePicker` now takes `disabled`, which closes the double-write and the out-of-order response that could persist a theme other than the one on screen.
- **Preserve a created connection when activation fails** — confirmed, including the slug detail: `create_connection` runs `uniquify`, so each retry left a `-2`, `-3` behind. The created id is retained and the retry activates it. The form is replaced by "…was created but could not be made active" rather than left editable — fields that are no longer submitted should not still look live — and the button says **Retry activation**.
- **Disable finishing while world creation is pending** — correct, and it is the same class as the two steps fixed last round, missed here. The failure ordering you describe is the bad one: `setup_done` written, navigation done, the only reporter of the result unmounted, so a failed creation leaves an empty library with setup permanently dismissed.
- **Recheck first-run state after changing stores** — this changes my answer from last round, and the narrower scope is why. I declined refreshing the verdict on *every navigation* because `finish()` treats the `setup_done` write as best-effort, so on a store that cannot record it the refetch says first-run again and skipping loops forever. A one-shot recheck after a *successful* data-dir move never touches that path. `StorageLocation` now reports `onMoved`, the wizard re-reads config, and pointing step one at an already-stocked library — the synced-folder case that step exists to support — turns step four into "this library is already stocked", with no create form and a live campaign handoff, instead of offering to add a stray world to an established collection.

My first-round rebuttal was sound about the mechanism but I generalised it into declining the whole concern, when the harm had a narrower fix. Thanks for re-raising it at the callback rather than at the route.

Each fix carries a test and I mutation-checked them: reverting all four fails exactly those four and nothing else. Full frontend suite is 60 files / 867 tests green, typecheck clean.

---

charlesmsiegel:
All seven of the third round are real and fixed in 4e2b4c2. Two of them correct this branch's own earlier fixes rather than the original design.

**Propagate permission failures from record probes** — the most important one, and it undercut an invariant stated in the docstring directly above the code. `is_dir()` and `exists()` answer `False` for a permission or I/O error exactly as they do for absence, so a library that merely could not be opened was reported empty, `_setup_state` never reached its fail-closed `except OSError`, and the verdict was `first_run: true` over an existing collection. `_names_a_record` now stats explicitly, and only `FileNotFoundError` counts as absence.

**Refresh first-run state after creating content outside the wizard** — third framing of this, and I'm taking it. My objection each time was that a live verdict traps a user whose `setup_done` write fails, since `finish()` is deliberately best-effort. That was an argument against refreshing *without a session latch*, not against refreshing — and I generalised it into declining the concern twice. Now:

```
firstRun   — live verdict, refreshed with the config on every navigation
leftSetup  — one-way latch, set when the wizard exits
inSetup    = firstRun && !leftSetup
```

Both routes branch on `inSetup`, so they remain exact complements and cannot bounce a redirect between them; the latch makes one exit always sufficient regardless of what the server answers. `createWorld` and `createCampaign` now invalidate the config cache too, since creating content changes `first_run` even though nothing wrote to `config.md` through the client — without that the refreshed read was still serving the pre-creation response, as you noted.

**The rest, all confirmed:**

- *Block setup skipping while a connection save is active* — same class as the two writes already guarded, missed again; `busy` is in the condition now.
- *Await the post-move first-run recheck* — the promise was discarded and `moving` cleared immediately, so Next re-enabled before the new store was classified. `onMoved` is awaited and counts as part of the move. On a failed read the wizard says so and leaves the create form up rather than guessing: claiming "already stocked" would strand a genuinely fresh user with no way to make a world.
- *Disable resetting storage during another move* — `saveDataDir` guards on `moving` and the Reset control is disabled, so pointer updates stay serialized.
- *Distinguish dismissed empty stores from stocked libraries* — correct, and the `CampaignWizard` consequence is real: its step 1 needs a world to select. The check is `listWorlds().length > 0` now, not `!first_run`.
- *Reset completed steps after changing stores* — `connected`, `createdId`, `worldId` and `worldName` are cleared on a successful move.

One note on process, since I claimed it in the previous two rounds: my first mutation check here failed only four of six reverted guards. The tests for the awaited recheck, the reset lock and the stocked check did not actually pin them — the last passed by accident because a default mock made the wrong expression coincidentally right. Those three have real tests now, and reverting each fails exactly its own case.

Gate: frontend 60 files / 873 tests green, typecheck and ruff clean, backend 3606 passed. The single backend failure is `test_atomic.py::test_a_read_only_record_is_not_silently_replaced`, which is pre-existing and environmental — it expects `PermissionError` from a `0444` file, and the dev container runs as uid 0; it reproduces identically on unmodified `main` and this branch does not touch `store/atomic.py`.

---

charlesmsiegel:
All four fixed in 1b1cf7f — and three of them share one root cause I had been routing around for three rounds.

**Serialize the setup backfill with config updates.** This is the finding that changes my position. `write_config` is an unlocked read-modify-write of the whole file; I identified that in round one, called it pre-existing, and deferred it to a separate issue — then spent rounds two, three and four disabling one more control each time a new frontend path to it turned up. That framing does not survive this comment: the writer that makes the loss reachable is the one *this branch adds*. `_setup_state` backfills `setup_done` from a **GET**, so a second tab merely loading the app races a setting saved in the first, with nobody doing two things at once. "Pre-existing" stopped being true the moment a writer went on the read path.

`store/config.py` now takes `locks.config_lock()` — a `_ProcessScopedLock` in the same style as the campaign and module-edit locks, so it excludes other processes and not just threads — around both the read-merge-write and the first-read materialization of the defaults. It is a leaf: nothing under it takes another lock, so it has no place in the ordering rules and cannot participate in a cycle.

**Fetch a fresh setup verdict on navigation** — correct, and the consequence is broader than the wizard: because the cache is only invalidated by this tab's writes, the connection-status pill in the topbar was equally stale for the life of the process. `getConfig` takes `{ fresh: true }` now and the per-navigation read uses it.

**Reload the selected store's theme after moving** — confirmed, including the second-order effect: the Theme step marked the previous store's card active, so clicking the card that looked already-selected would have written that value into the new library's config.

**Lock the wizard while finishing** — confirmed. Every control now holds on a single `writing` flag covering all four in-flight writes, replacing the four ad-hoc conditions that kept missing one; `finish()` also guards re-entry, so the two destinations cannot both fire.

One process note, since it is the second round running where my testing was the weak link: the first version of the lock test **passed with the lock removed**. It drove both writers through `TestClient`, which serializes requests across threads, so it interleaved nothing and asserted a race it never created. It drives `write_config` directly now, and removing the lock fails it.

Gate: frontend 60 files / 876 tests, backend 3607 passed, typecheck and ruff clean. The one backend failure remains `test_atomic.py::test_a_read_only_record_is_not_silently_replaced` — pre-existing and environmental (expects `PermissionError` from a `0444` file; the container runs as uid 0), reproduces identically on unmodified `main`.

---

charlesmsiegel:
All four of the fifth round are fixed in 3070061. Two of them correct guarantees this PR's own description claimed.

**Persist setup history when content is created.** Correct, and it falsifies something I wrote: the docstring said "deleting every world later does not bring the wizard back", but the flag was only recorded when a config read happened to find content. Create a world and delete it on the Worlds page before anything reads the config — no trace, and the next navigation reopens setup. `config.mark_setup_done()` is now called when the first world or campaign is created, so the guarantee no longer depends on a read interleaving. The read-side backfill stays for stores that predate the key.

**Reset the setup-exit latch after switching stores.** Also correct, and it is a defect in a fix I added two rounds ago for the opposite hazard. The latch existed so a failed `setup_done` write could not re-trap a user; being one-way and unscoped, it also suppressed the wizard for every *other* store the session touched, including a brand-new empty library chosen from Config. `_public_config` now reports the `data_dir` it describes, `finish()` passes back the store its answer belongs to, and the latch is keyed on it — same protection, correct scope.

**Restore an active connection when setup resumes.** Confirmed, including the duplicate: re-entering the form calls `createConnection` again and the slug uniquifies. The step adopts the store's active connection now, gated on `ready` rather than mere presence — a fresh store ships with an OpenRouter connection selected and no key, so the weaker check would have declared every genuinely-fresh install "connected" and skipped the step that exists to fix exactly that.

**Revert the theme preview when persistence fails.** Confirmed. Left applied it looks chosen for the session and disappears on reload, which reads as the app losing a setting rather than never accepting it.

Each has a test; reverting all four fails exactly those four.

I also merged `origin/main` (36 commits ahead, across five files this branch touches — `store/locks.py` most consequentially, where this branch adds `config_lock`). Clean merge, and the post-merge suite was green before these fixes went in.

Separately: I have corrected the PR description. It still carried a note saying the Codex gates could not run because the CLI was not installed in the authoring container — that was wrong, and leaving it would have told a reviewer the gate never ran when in fact it ran on every push and found 24 issues. The verification table was also stale by four commits.

Gate on the merged branch: backend 3860 passed, frontend 60 files / 898 tests, lint clean, templates 62 checks. The single backend failure remains the pre-existing `test_atomic.py` one that expects `PermissionError` from a `0444` file while the container runs as uid 0.

---

PR #329, opened 2026-08-14
Keep /api a 404 when the path is spelled with backslashes

### LINKED ISSUE #313: SPA fallback answers unknown /api paths with index.html on Windows
## What happens

`backend/tests/test_spa_fallback.py::test_an_unknown_api_path_stays_a_404_even_for_a_browser` fails on Windows. A navigation to a mistyped API path gets `index.html` with a **200** instead of a 404.

```
assert r.status_code == 404
E       assert 200 == 404
```

## Why

`SPAStaticFiles.get_response` (`backend/src/grimoire/main.py:65`) guards `/api` by string prefix:

```python
if exc.status_code != 404 or path == api or path.startswith(api/):
    raise
```

`path` comes from Starlette''s `StaticFiles.get_path`, which runs `os.path.normpath`. On Windows that returns `api\not-a-real-endpoint`, so `startswith(api/)` is never true and the request falls through to the `index.html` fallback. (`path == api` still matches, so only nested paths are affected — which is every real endpoint.)

The `text/html` check on the next line is intact; this is purely the separator.

## Why CI is green

The GitHub Actions runners are Ubuntu, where `normpath` keeps forward slashes and the guard works. The failure only appears on a Windows checkout — and Windows is a first-class target here, since the packaged desktop build and the Android WebView shell both serve through this exact class rather than `vite dev`.

## Impact

An unknown or mistyped `/api/...` endpoint returns 200 + an HTML document instead of a 404. Any client that checks the status code before parsing sees success and then fails on the body.

## Suggested fix

Normalize the separator before comparing, e.g. `PurePath(path).parts[0] == api`, or `path.replace(os.sep, /)` first. Worth a test that exercises the Windows separator explicitly so Ubuntu CI can still catch a regression.

## Origin

Introduced with #299 and merged unfixed — it reproduces at that PR''s original branch tip, so it was not a rebase artifact. Flagged on the PR at the time (#299 review comment) but the PR landed before it was addressed.

charlesmsiegel:
Fixes #313.

## The bug

`SPAStaticFiles.get_response` guarded `/api` by string prefix:

```python
if exc.status_code != 404 or path == "api" or path.startswith("api/"):
```

`path` comes from Starlette's `StaticFiles.get_path`, which is
`os.path.normpath(os.path.join(*route_path.split("/")))`. On Windows that hands
back `api\not-a-real-endpoint`, so `startswith("api/")` never fires and the
request falls through to the `index.html` fallback: an unknown endpoint answered
**200 with an HTML document** instead of a 404. `path == "api"` still matched, so
only *nested* paths were affected — which is every real endpoint. A client that
checks the status before parsing sees success and then fails on the body.

Not a developer-only surface: the packaged desktop build and the Android WebView
shell both serve through this class rather than `vite dev`.

## The fix

`_under_api` normalizes the separator before comparing, rather than trusting the
platform's — Ubuntu CI cannot produce the failing one, so a fix keyed to `os.sep`
would be correct and untestable. It compares the whole first segment, so
`/apiary` and `/api-docs` keep their SPA fallback. The `text/html` check on the
next line is untouched.

It reads the *resolved* path rather than the raw `scope["path"]`, which would
sidestep the separator question entirely and be worse: `get_path` has already
collapsed `..`, so `/x/../api/gone` arrives as `api/gone`, and a check against
the URL would read the literal spelling and hand back the SPA for a request the
file lookup had already resolved into `/api`.

## Tests

`_windows_get_path` is Starlette's own `get_path` with `ntpath` substituted for
`os.path` — the same resolution Windows performs — monkeypatched onto the class,
so the regression is reproducible on Ubuntu. Both directions are covered: an
`/api` path keeps its 404, and a nested client route keeps its fallback.

Two things back that up, because a simulation is only as true as its copy:

- `_under_api` has direct tests, including the bug report's path spelled
  verbatim. Every other test reaches the predicate through the simulation — if
  the simulation's premise is wrong, those go green while Windows stays broken.
  These cannot.
- The simulation itself is asserted equal to Starlette's real `get_path` on
  whatever platform is running. If `get_path` grows a step the copy does not
  model (`root_path` stripping, a different join), that fails loudly instead of
  quietly simulating something Starlette no longer does. Verified by drifting
  the copy on purpose.

## Also here: three test-only frontend commits

Unrelated to #313 — this PR's `frontend` job failed on `CampaignView.test.tsx`,
and two of the sites turned out to be measuring the wrong thing rather than
merely failing at random. Each is mutation-tested. See the thread for the
measurement: that file fails an arbitrary test in roughly two thirds of
full-suite runs, with these commits and without them, so **they do not make the
frontend job reliable** — that is a separate, larger piece of work.

## Verification

- `make check-py` — 4653 passed, 1 failed: `test_atomic.py::test_a_read_only_record_is_not_silently_replaced`,
  which fails identically on the unmodified tree in this container. It runs as
  root, where `chmod 0o444` does not stop a write; unrelated to this change.
- `make check-lint` (ruff), `make check-templates` (81 checks), `tsc -b` — clean.
- The real `create_app()` was driven end-to-end for both separators, the
  percent-encoded form, and the near misses.
- Not run locally: `check-pydantic1`. Not tested: an actual Windows host — the
  fix rests on the simulated separator plus the string in the bug report.

The repo's Codex review gates were not run — Codex is out of usage. A self-review
pass stood in for them; its findings and verdict are in the thread.

https://claude.ai/code/session_01MtjRXqvT64QcrZgDP8ojYM


charlesmsiegel:
Two follow-up commits.

**`frontend` job failure — not this PR's change, but fixed here.** The job failed on `CampaignView.test.tsx` → "typed guidance is passed to regenerate": `getByTitle("Reroll")` found nothing after awaiting the reply the control hangs off. That is the same mechanism cf46b0c fixed for the response picker — `findByText("old reply")` waits on the transcript chain, while the Reroll button additionally needs `transcriptIsActive`, which `selectScene` establishes on a different promise chain; two chains are two React commits, and `findBy` polls on a timer. cf46b0c fixed the two sites it had proof for and explicitly left the rest rather than rewriting on suspicion, noting the failure lands on an arbitrary member of the group each run. CI has now supplied the proof for this selector, so the whole `getByTitle("Reroll")`-immediately-after-an-awaited-message cluster moves rather than only the one that lost this particular race.

I could not reproduce it locally — the full suite passes here under coverage (74 files, 1283 tests), which is consistent with it being scheduling-dependent. So the claim is not "verified cured"; it is that the remedy is the one already established for this mechanism, applied to a site CI proved is subject to it. Happy to split it into its own PR if you would rather keep this one to the backend fix.

**Self-review of the #313 fix.** Codex is out of usage, so I reviewed it myself and found one thing worth acting on: every test for the fix reached the predicate through `_windows_get_path`, which is only this repo's *belief* about how Windows resolves a URL. If the belief is wrong, the tests go green and Windows stays broken. `_under_api` now has direct tests, including the bug report's path spelled verbatim, plus the near misses a sloppier fix would have broken (`/apiary`, `/api-docs` — `startswith("api")` would have cost those client routes their fallback).

One criticism I am consciously **not** acting on: the API mount point is now written down in three unconnected places in `main.py` — `include_router(prefix="/api")`, `_under_api`'s `"api"`, and `_CampaignActivityStamp._PREFIX`. One of those three just shipped a bug by drifting from the routing. A shared constant would be the honest fix, but `_PREFIX` carries its own documented reasoning and rewriting it is churn beyond this issue. Flagging it rather than bundling it — say the word and I will do it.

---

charlesmsiegel:
**Correction to my last comment, and a measurement.** I said the frontend failure did not reproduce locally. That was true of the two runs I had done and false in general — it reproduces readily. Ten full-suite runs on this machine, half of them with my frontend changes stashed:

| | runs | failed | where |
|---|---|---|---|
| with my changes | 7 | 4 | `CampaignView.test.tsx`, always |
| changes stashed | 3 | 2 | `CampaignView.test.tsx`, always |

Roughly sixteen distinct test names across those failures, never the same one twice, every single one an "Unable to find an element" immediately after an `await`. Two runs failed while nothing else was on the box, so it is not purely contention — it just gets more likely under load, which is what a 2-core CI runner is.

So the honest picture, replacing what I said before: **this is a property of `CampaignView.test.tsx`, not of any selector in it.** cf46b0c named the mechanism — a scene selection fans out a dozen independent promises, `findBy` polls on a timer, and any `getBy`/`queryBy` sitting immediately after a single `await` is a coin flip on which commit the poll landed in. There are hundreds of those sites in the file. Fixing them one CI failure at a time cannot converge, and my three commits here do not make the frontend job reliable — every PR against this repo will keep drawing a random red job until the file is addressed systematically. That is its own piece of work, and I have not attempted it here.

What I did keep is the sites with evidence behind them, each mutation-tested:

- the `Reroll` waits (CI proved that selector), and
- two assertions that were measuring the wrong thing rather than merely failing at random: `no Reroll on an all-assistant transcript` asserted an *absence* before the gutter had rendered — it would have passed while the guard it tests was broken — and `a complete scene replaces the composer with a notice` read the notice flat, which is the test that failed locally. Both now anchor on the state they mean to measure; removing the all-assistant guard, and rendering the composer unconditionally, each still fail them.

**On the backend fix itself**, one more review pass found one thing worth acting on: `_windows_get_path` is a hand-copy of Starlette's `get_path`, and the two tests that reproduce #313 are only as true as that copy. If `get_path` ever grows a step the copy does not model, both keep passing while simulating something Starlette no longer does. The copy is now checked against the real method on whatever platform is running — verified by drifting it on purpose.

`make check-py`: 4653 passed, 1 failed — `test_atomic.py::test_a_read_only_record_is_not_silently_replaced`, which fails identically on the unmodified tree in this container because it runs as root, where `chmod 0o444` does not stop a write.

---

PR #331, opened 2026-08-14
fix(security): pin avatar-download socket to the validated IP to close the DNS-rebinding gap

### LINKED ISSUE #28: fix(security): pin avatar-download socket to the validated IP to close the DNS-rebinding gap
## Context
Avatar download (and card-text image localization) fetch remote images through `store/fetch.py`, which has a real but TOCTOU-vulnerable SSRF guard. The flow: `characters._download_avatar` → `fetch.download_url` → `fetch._http_get_bytes` (`characters.py:235-249`, `fetch.py:107-122`, `fetch.py:77-104`).

The existing guard (`fetch._http_get_bytes`, `fetch.py:77-104`) does the following per request and **per redirect hop** (manual redirect loop, `follow_redirects=False`, up to `_MAX_REDIRECTS = 5`, `fetch.py:23`, `:80-93`):
1. `urlparse` the URL and reject non-`http(s)` schemes / missing hostname (`fetch.py:82-84`).
2. Call `host_is_blocked(hostname)` (`fetch.py:85`), which `socket.getaddrinfo`s the host and blocks if **any** resolved address is private, loopback, link-local, reserved, multicast, or unspecified — and blocks unresolvable hosts (`fetch.py:60-74`).
3. Only then `client.stream("GET", url)` (`fetch.py:87`).

The residual risk: `host_is_blocked` resolves the name, but then httpx **re-resolves the same hostname when it opens the connection**. A DNS-rebinding attacker who controls the authoritative DNS can return a public IP for the guard's `getaddrinfo` and a private/loopback IP (e.g. `169.254.169.254`, `127.0.0.1`) for httpx's connect — a classic check-then-connect TOCTOU. There is also a size cap (`MAX_BYTES = 100 MiB`, `fetch.py:19`, `:96-102`) and magic-byte validation (`fetch.py:115-117`), but those do not address the rebinding window. The TLS context trusts certifi explicitly (`fetch.py:26`, `:80`).

## Desired behavior
Connect only to the exact IP that was validated, eliminating the re-resolution window, while preserving correct TLS (SNI + certificate hostname verification) against the original hostname.

## Implementation notes
- Resolve the hostname **once** per hop, validate every returned address with the existing predicate logic (reuse/refactor `host_is_blocked`, `fetch.py:60-74`), then pin the connection to a chosen validated IP.
- With httpx, pin via request extensions rather than re-resolving:
  - Build the request URL using the **literal validated IP** as the host (and original port), set the `Host` header to the original hostname, and pass `extensions={"sni_hostname": <original-host>}` so httpx uses the real hostname for both TLS SNI and certificate verification while connecting to the pinned IP. This keeps `verify=_SSL_CTX` (`fetch.py:80`) meaningful.
  - Apply the same pinning on each redirect hop (the loop already re-validates the new URL's host at `fetch.py:82-85`; extend it to pin the resolved IP for that hop too).
- Alternative if extension-based SNI proves insufficient: a custom `httpx.HTTPTransport`/transport map that forces the connection target to the pre-validated IP for the request's host, leaving URL/Host/TLS hostname intact.
- Keep behavior best-effort: any resolution/validation failure raises inside the `try` and `download_url` returns `None` (`fetch.py:109-112`) — no caller should start failing.
- Refactor so `host_is_blocked` can return the validated address(es) (e.g. a `resolve_allowed(host) -> list[str] | None`) to avoid resolving twice.
- IPv6: ensure the literal-IP URL form is bracketed.

## Acceptance criteria
- [ ] A hostname that resolves to a public IP at validation time but to a loopback/link-local/private IP at connect time cannot cause a connection to the private IP (covered by a test that stubs resolution to differ between the guard and the connect).
- [ ] Legitimate public HTTPS image downloads still succeed with valid certificate verification against the original hostname (SNI correct).
- [ ] Redirects to a blocked host are still rejected at each hop, and each hop connects only to its validated IP.
- [ ] The size cap, magic-byte validation, and best-effort `None`-on-failure contract are preserved.
- [ ] Backend test simulates the rebinding split-resolution and asserts the private connect is blocked.

## Source
TODO.md §Character card polish — deferred scope / known limitations

charlesmsiegel:
Closes #28.

> **Note on this PR's diff**: the branch was handed to me already carrying two
> unrelated commits (`129e916` frontend redesign, `cf46b0c`) that are not on
> `main`, so they appear in this PR. Only `backend/src/grimoire/store/fetch.py`
> and `backend/tests/test_fetch_store.py` are part of the issue-28 fix. Say the
> word and I'll re-cut the fix onto a branch off `main` so this PR is just the
> two files.

## The gap

`fetch.host_is_blocked` resolved the hostname and returned a verdict; httpx
then resolved the same name again when it opened the connection. An attacker
who controls the authoritative DNS answers those two lookups differently —
public for the check, `127.0.0.1` / `169.254.169.254` for the connect — so the
guard never saw the address the socket actually reached.

## The fix

- **`resolve_allowed(host) -> list[str] | None`** replaces the boolean core:
  one `getaddrinfo`, the same private/loopback/link-local/reserved/multicast/
  unspecified predicate over every returned address, and the validated
  addresses come back instead of a verdict. `host_is_blocked` stays as a façade
  over it, so nothing resolves twice.
- **The request is pinned to a validated literal.** The IP becomes the URL host
  (httpx brackets IPv6 itself), the original hostname stays in `Host` and in
  `extensions={"sni_hostname": ...}` — so TLS still does SNI and
  certificate-hostname verification against the real name, and `verify=_SSL_CTX`
  keeps meaning what it says. This holds through a proxy too: the literal is
  what gets CONNECTed, so the proxy doesn't re-resolve either.
- **One parser decides what the host is.** `urlparse` picking the name to
  validate while httpx picks the name to request is its own bypass — the two
  disagree about IDN hosts — so `httpx.URL` now does both, and the SNI hostname
  is the A-label (`raw_host`). Passing the unicode form would have handed `ssl`
  a name it re-encodes with IDNA 2003: for `faß.example` that is `fass.example`
  against httpx's `xn--fa-hia.example`, i.e. verifying a different site's
  certificate.
- **Every redirect hop resolves and pins the same way**, and a relative
  `Location` is joined against the hostname URL rather than the pinned one, so
  the next hop isn't addressed to a bare IP.
- **Keep-alive off.** httpcore's `can_handle_request` compares origin only, and
  the origin is now an IP; a redirect to a different hostname on the same
  address would otherwise reuse a connection whose handshake named the old one.
- **Multi-address hosts fall back**: a `ConnectError`/`ConnectTimeout` on one
  validated address tries the next, so pinning doesn't cost the resilience
  `getaddrinfo` ordering used to provide (e.g. AAAA-first on a v4-only box).

A proxy that allowlists hostnames will refuse an IP `CONNECT`; that fails
closed — the caller sees the usual `None` — which is the right way round for a
guard.

The size cap, magic-byte validation and the best-effort `None`-on-failure
contract are untouched; no caller's signature changed (`_http_get_bytes` gained
a keyword-only `transport=` test seam).

## Tests

`backend/tests/test_fetch_store.py`, 35 passing:

- **The rebinding case over a real socket** — resolution answers public once and
  loopback thereafter; the fetch is asserted to look the name up *exactly once*
  and to connect by the validated literal.
- **A blocked first hop sends nothing**, across nine spellings including the
  obfuscated loopback forms `2130706433` and `0x7f.0.0.1` — which the guard
  catches because it resolves rather than pattern-matches the URL text.
- Pinning at the request level: validated IP as URL host, hostname in `Host` and
  `sni_hostname`, port/path/query preserved, IPv6 bracketed, IDN host resolved
  and verified under one label.
- Per-hop behavior: each hop pinned to its own validated address; a redirect to
  a link-local host rejected before that hop is ever sent; a relative redirect
  keeping the hostname.
- **Against a local TLS server**: a certificate issued to the URL's hostname is
  accepted, one issued to another name is rejected with a certificate error —
  both while connected to a pinned literal. Skips where no `openssl` can issue
  a test cert.
- Multi-address fallback and the preserved size cap.

Verified by mutation, not by eye: un-pinning the request fails 10 tests,
validating only the first hop fails 1, dropping the `sni_hostname` extension
fails 4 — the TLS test among them, so it is load-bearing rather than
decorative.

Full backend suite: 4659 passed. One unrelated pre-existing failure,
`test_atomic.py::test_a_read_only_record_is_not_silently_replaced`, which
asserts a `PermissionError` that a uid-0 sandbox never raises.

## Note on the review gates

CLAUDE.md's implementation→done gate calls for `/codex:review`. Codex was
unavailable, so an adversarial self-review stood in for it; the second commit
is what that review turned up.


charlesmsiegel:
Closing in favor of #335, which carries the same fix cut fresh from `main` — two files instead of 93.

This branch was handed to me already carrying `129e916` (frontend redesign) and `cf46b0c`, neither of which is on `main`, so they showed up in this PR's diff alongside the security fix. I haven't touched the branch: both commits are still on `origin/claude/github-issue-28-h1x1zm`, and reopening this PR is one click if you want them reviewed on their own.

---

PR #333, opened 2026-08-14
Name every image from its bytes, not from the filename that claimed it

### LINKED ISSUE #321: Record images take their EPUB manifest media type from the filename, so a mislabelled upload produces an invalid book
Every image packed into an exported EPUB gets its `media-type` from the stored file's suffix, and that suffix comes from whatever the uploader's filename said. Nothing checks the bytes. A JPEG uploaded as `avatar.png` is stored as `avatar.png`, packed as `images/img-000.png`, and declared `media-type="image/png"` in `package.opf` — which epubcheck reports as an error, and which some readers refuse to render.

Campaign covers were fixed for this in the cover-image work (`store/covers.py` derives the extension from PIL's detected format and rejects anything outside the allowlist). Record images — character avatars and galleries, greeting art, location and lore images — were deliberately left alone there, because the hazard is pre-existing, reaches every image route, and affects stores already on disk. `_FORMAT_EXT` in `store/covers.py` records that scope decision. This issue is the other half.

## Reproduction

Against `main`:

```python
import io, zipfile, re, tempfile, os
os.environ['GRIMOIRE_HOME'] = tempfile.mkdtemp()
from PIL import Image
from grimoire.store import worlds, campaigns, characters, assets, scenes, appearances, epub

wid = worlds.create_world('Realm'); wroot = worlds.world_root(wid)
ch, _ = characters.create_character(wroot, 'Mara', 'default',
                                    {'spec': 'chara_card_v3', 'data': {'name': 'Mara', 'description': 'x'}})

buf = io.BytesIO(); Image.new('RGB', (40, 40)).save(buf, 'JPEG')
assets.put_image(wroot, ch, 'default', 'avatar', buf.getvalue(), 'png')   # JPEG bytes, .png name

cid = campaigns.create_campaign('Saltmarch Nights', wid)
sid = scenes.create_scene(cid, 'Arrival')
appearances.appear(cid, sid, 'characters', ch, 'default', 'npc')

z = zipfile.ZipFile(io.BytesIO(epub.build_epub(cid)[0]))
img = [n for n in z.namelist() if n.startswith('images/')][0]
declared = re.search(rf'href="{re.escape(img)}" media-type="([^"]+)"', z.read('package.opf').decode()).group(1)
print(img, declared, Image.open(io.BytesIO(z.read(img))).format)
```

Output:

```
images/img-000.png image/png JPEG
```

## Where it comes from

- `routes/characters.py:379`, `routes/entities.py:190`, `routes/campaigns.py:565` — the stored extension is `filename.rsplit(".", 1)[-1]`, validated only against the allowlist, never against the bytes.
- `store/export.py` `Images.add` — the packed name reuses `p.suffix.lower()`.
- `store/epub.py:127` — `media_type` is `_EXT_MEDIA.get(name.rsplit(".", 1)[-1], ...)`.

`build_html` has a milder version of the same thing: its base64 data URIs carry the same wrong MIME, which browsers survive by sniffing.

## Worth deciding

The interesting part isn't the fix, it's what to do about stores that already hold mislabelled files, since the extension is part of the filename and renaming one changes the URL every existing reference uses:

1. **Validate on upload only.** New uploads get the extension from the detected format, exactly as covers do. Existing bad files stay bad, and books exported from them stay invalid.
2. **Also fix at pack time.** The exporter sniffs each image and declares the true media type regardless of the name — the book is valid even from an old store, and nothing on disk moves. The declared type then disagrees with the packed filename's suffix, which is legal but odd.
3. **Both, plus a repair pass** that renames mislabelled files in place and repoints the markdown image URLs that reference them (`store/localize.py` writes those URLs; `scene_refs.repoint()` is the nearest existing precedent for a rewrite of this shape).

My inclination is 1 + 2: upload validation stops the bleeding, pack-time sniffing makes every export valid immediately, and the on-disk rename can stay unwritten until something else needs it. Worth noting that 2 alone is the only option that fixes books exported from stores that already exist.

Related: #178 (cover-image auto-generation for exports) — campaign covers now exist as a manual upload, so that issue is narrower than when it was filed.

charlesmsiegel:
Closes #321.

Takes the issue's own recommendation — **1 + 2**: validate on upload, *and* name at pack time. No on-disk repair pass (option 3); nothing in the store is renamed.

## The defect

Record images took their stored extension from the uploader's filename, and every consumer then declared a media type from that extension. A JPEG uploaded as `avatar.png` was stored as `avatar.png`, packed as `images/img-000.png` and declared `media-type="image/png"` in `package.opf` — an epubcheck error, and a book some readers refuse to render. The extension allowlist never closed this: it was only ever checked against the filename, which is the thing that lied.

The issue's reproduction, against this branch:

```
images/img-000.jpg image/jpeg JPEG
```

## The two halves

**Upload validation** — `routes.common._upload_image_ext` names the stored extension from the bytes' signature, for all three image `PUT` routes (world character, campaign character, entity). Bytes in no format we can label are a `400` rather than a file stored under a name that lies about them; covers have refused those since `covers.validate`.

**Pack-time naming, and dropping what cannot be named** — `export.Images` names every packed image from its own bytes, so a book exported from a store that *already* holds mislabelled files is valid too. That is the half upload validation cannot reach, and per the issue it is the only option that fixes stores that already exist. The packed name is generated (`img-000`), so the name and the declared type agree rather than packing `img-000.png` and declaring `image/jpeg`.

An image whose bytes name **no** format we can declare is not packed at all. It degrades to its alt text — the path `rewrite_images` already had for a remote or missing image — and an actor whose avatar can't be named gets the no-portrait rendering the appendix already had. The EPUB cover takes the same route into the well-tested no-cover book. The manifest therefore declares a core EPUB image type for every entry it has, by construction.

The store keeps the file and the app keeps serving it: a response has to carry some type and a browser sniffs past a wrong one. Only the book declines, because an EPUB reader validates the manifest and may refuse the image outright — and a refused image shows the same nothing the alt text shows, minus the caption.

That is what makes it safe to leave `fetch.download_url`'s last-resort `ext = "png"` alone. An AVIF avatar it names `.png` still downloads, still renders in the app, and simply isn't carried into a book that would be invalid for carrying it. Dropping the download instead would have lost the image outright.

The same rule closes the other doors that read the stored suffix, each a one-line change on the value already in hand:

- serving an image sets its `Content-Type` from the bytes, falling back to the suffix (`_serve_image_file`);
- an exported card's `data:` mime and its CHARX member name come from the bytes (`characters._stored_avatar`);
- `fetch.decode_data_uri` believes the bytes over the URI's declared mime — a JPEG embedded as `data:image/png;base64,…` was the same defect arriving through card import and localization.

Incidental robustness: an unreadable image (permissions, a directory in its place) used to register at collect and raise at zip time, failing the whole export. It is now simply absent from the book.

## Why magic bytes rather than covers' PIL decode

The export names every packed image with the same detector the upload gate uses, and using one at both ends is what makes a stored image's suffix the suffix the packer derives, rather than two rules that merely agree today. They differ only in what they do with "no format I can name": the upload is refused, the image is left out of the book. Covers stay PIL-based because only the decode bounds the raster they thumbnail (`MAX_PIXELS`); all three map the same four formats to the same four extensions.

Record image uploads deliberately do **not** gain covers' size and pixel caps — that is a policy change this issue did not ask for.

## Frontend

The image file pickers accept exactly those four types instead of `image/*`, so the dialog no longer offers files the server will now reject. The upload handlers already surface `err.detail` in their error banner, so the `400` renders as "unsupported image type".

## Known, and deliberate

- **A narrow race remains, reproduced not theorized.** The header is read while the export is composed and the bytes are packed later, so an external writer that swaps a file's *format* in place under the same name mid-export still gets one wrongly-declared image. No app path can do it — an upload of another format lands on another suffix and drops the old file, failing the pack loudly instead. Closing it costs every image held in memory from collect to zip, or a manifest rebuilt at zip time. Documented on `Images`.
- **Option 3 would be cheaper than #321 estimated**, worth knowing before that issue is scheduled. The premise was that renaming a file changes the URLs referencing it; it doesn't. App image URLs address the *logical* name (`…/images/avatar`); the extension exists only in the filename `assets.image_path` globs for. A repair pass would rename files and repoint nothing.
- **Not verified against epubcheck** — not available in this environment. Verified as "the declared type matches the bytes, and every manifest entry is a core image type".

## Tests

- The issue's exact repro as a regression test (`test_epub_store.py`), plus the cover.
- All four formats the store accepts, uploaded under names that lie about them, checked through the response, the listing, the served `Content-Type` and the file on disk (`test_routes.py`) — which also walks `put_in`'s drop-the-stale-sibling path across four suffixes.
- Non-image bytes are a 400 on all three routes; a misnamed file already on disk serves as what it is.
- Dropping: an unnamable image degrades to alt text and stays in the store; an unnamable cover produces the no-cover book; an unnamable portrait leaves the actor without one; an unreadable image no longer fails the export.
- Card export names the avatar from its bytes in JSON and CHARX; `decode_data_uri` believes the bytes.
- Fixtures that stored marker strings as images (`b"pierbytes"`, `b"face"`, `b"old"`/`b"new"`) are real images now — under a rule that reads bytes, those were never images, and the tests only passed because nothing looked.

Verified by reverting `backend/src` to `main` with the new tests in place: each one fails without the fix.

`make check` passes locally — pytest (4648), ruff, `verify_templates.py`, the pydantic-1.10/Android dependency set, and the frontend typecheck + vitest (1283).

https://claude.ai/code/session_018pbPg9hMFC5ZtoSu3mzzBZ


charlesmsiegel:
The `frontend` job is red on `a30abef`, and I don't believe it belongs to this branch. Recording why rather than pushing at it.

`a568d2a` → `a30abef` changes two files, both backend (`store/export.py`, `tests/test_routes.py`); `git diff --name-only a568d2a a30abef -- frontend/` is empty. The `frontend` job passed on `a568d2a` and failed on `a30abef` with byte-identical frontend sources and the same `npm ci` lockfile — same input, different result.

Three *different* tests in `src/routes/CampaignView.test.tsx` have failed across the runs so far, each an async/timing assertion:

- `normal scene: empty Continue sends an ephemeral round` — spy never called
- `HTML comments are invisible in the rendered transcript but survive into the edit box` — `Unable to find an element with the title: Edit message`
- (locally, once, under full-suite load) the `next reply only` accessible-description assertion

All three pass in isolation and on re-run; the whole file passes locally at `HEAD` and at `HEAD~1`. `main` shows the same intermittency — 3 failures in its last 12 runs — and the current `main` HEAD is *"Wait for the response picker's value, don't assert it flat"*, which is this same class of assertion being stabilized.

This branch's only frontend change is narrowing `accept="image/*"` to the four types the server now accepts, on two file inputs in `CharacterEditor`/`EntityEditor`. `CampaignView` renders neither.

I'll re-run the job once the workflow finishes and keep watching. If it keeps landing on this file, the fix is stabilizing those assertions in `CampaignView.test.tsx` — a separate change from this one, and I'd rather not bury it in an image-media-type PR.

---

Issue #31, opened 2026-07-11
Reversible change/delta log with true undo of writes


charlesmsiegel:
Part of the **Storage & Configuration** milestone. From the pre-rebuild target feature list:

> Reversible change/delta log with true undo of writes

## Current state

`backend/src/grimoire/store/changes.py` keeps a *rolling* per-record change log at `<campaign>/changes.json`, keyed `"<kind>/<id>"` → `{scene, fields: [{field, label, before, after}]}`. `record()` upserts — only the **latest** write-back per record survives. It is written from exactly one place: `absorb.apply_edits(cid, edits, sid)` in `store/absorb.py`, and only for `_BROWSABLE_KINDS = ("character_state", "lore", "authored")`; relationship/bond/plot applications are never recorded. Read side: `GET /api/campaigns/{cid}/changes` (`routes.py:get_changes`) + `changes.line_diff()`, rendered by `frontend/src/components/ChangesPanel.tsx`.

So the before/after needed for undo already exists per edit — but there is no history (each write-back clobbers the last delta), no coverage of manual PUT edits (entity/character routes write directly), and no undo endpoint anywhere. `docs/superpowers/PHASE-STATUS.md` lists "only the **last** write-back delta per record (rolling)" as a known deferral.

### Option A — append-only journal + inverse-apply (recommended)
- New `<campaign>/journal.json` (append-only list): `{ts, scene, ref, kind, field, before, after, source: "absorb"|"manual", undone?}`.
- Append from `absorb.apply_edits` (all kinds, including relationship/bond/plot payloads) and optionally from campaign-scope mutating routes in `routes.py`.
- `POST /api/campaigns/{cid}/journal/{n}/undo`: write `before` back through the same writers `apply_edits` uses (`playstate.write_state`, `entities.update_entity`, `characters.update_version` via `appearances.locked_version`, `relationships.set_feeling/set_bond`). **Guard:** apply only when the current value still equals `after`, else 409 — a later edit intervened.
- Keep `changes.json` for ChangesPanel or derive it from the journal tail.
- Tradeoff: touches every write path, but yields field-precise, auditable, conflict-safe undo.

### Option B — pre-apply file snapshots
- Before `apply_edits` writes, copy each target file into `<campaign>/.undo/<sid>/`; undo restores files wholesale.
- Exact and trivially correct; but whole-file granularity (undoing one edit reverts unrelated same-file edits from the same absorb) and no manual-edit coverage.

[...]

PR #341, opened 2026-08-14
Cascade post-delete with state reversal (#75)

### LINKED ISSUE #75: Cascade post-delete (delete a post and everything after it) with state reversal
Part of the **Scenes & Play** milestone. From the pre-rebuild target feature list:

> Cascade post-delete (delete a post and everything after it) with state reversal

## Current state

`backend/src/grimoire/store/scenes.py` has two message-mutation primitives today: `edit_message(cid, sid, index, content)` (in-place content edit, any index) and `remove_trailing_assistant_run(cid, sid)` (pops only the trailing run of assistant messages — one turn's output, not an arbitrary cut point). There is no "delete this post and everything from here on" primitive, and no route (`routes.py` scene routes stop at `PUT .../messages/{index}`, `put_scene_message`).

"State reversal" is intrinsically bounded by what the continuity pipeline actually tags with a scene id. Canonical writes only happen at end-of-scene review (`absorb.apply_edits`, invoked by `PUT /campaigns/{cid}/scenes/{sid}/chronicle`) — not per post. Of the stores `scene_refs.repoint` fans out to (`appearances`, `chronicle`, `changes`, `plot` — see `store/scene_refs.py`), only `chronicle.json` (keyed by scene id, one record per scene) and `plot.json` (`beats[].scene`) and `changes.json` (`{"scene": sid, "fields": [...]}`, but **rolling: only the latest write-back per record is kept**, `store/changes.py:record`) carry a scene tag. `relationships.json` (`store/relationships.py`) has **no scene attribution at all** — bonds have a `since_scene` field but it is schema-present and never populated (per `docs/superpowers/PHASE-STATUS.md`), and `relationships` is conspicuously absent from `scene_refs.repoint`'s fan-out list. So a cascade delete can reliably undo the scene's chronicle record, its plot beats, and its changes.json rows, but cannot safely revert relationship feelings/bonds it may have caused.

### Option A — Truncate + scene-scoped rollback of only what's attributable (recommended)
- Add `scenes.delete_from(cid, sid, index)`: slice `messages[:index]`, re-serialize (mirrors `remove_trailing_assistant_run`'s read→slice→`_serialize_messages`→write shape).
- Route `DELETE /campaigns/{cid}/scenes/{sid}/messages/{index}` (mirror `put_scene_message`), 400 on out-of-range index.
- If the scene is already absorbed (`meta["done"] == "true"`) and the cut removes absorbed content: delete the scene's `chronicle.json` record, drop `plot.json` beats where `beat["scene"] == sid`, drop `changes.json` rows where `rec["scene"] == sid`, clear `meta["done"]`/`one_line`/`summary` so absorb can re-run. Leave relationships untouched and say so explicitly in the UI copy ("relationship changes from this scene are not reverted").
- Frontend: a delete icon in the post gutter (`frontend/src/routes/CampaignView.tsx`, near the existing `✎`/reroll buttons in `.msg-gutter`) with a confirm dialog stating the post count removed and which continuity records get reverted.

### Option B — Reapply changes.json's before-value instead of just deleting rows
- More faithful for `character_state`/`lore`/`authored` edits (the `_BROWSABLE_KINDS` in `absorb.py`): write `rec["fields"]`'s `before` back onto the live record via `playstate.write_state`/`entities.update_entity`/`characters.update_version`, rather than merely dropping the changes.json row.
- Tradeoff: `changes.json` only stores the *last* write-back per record (`record()` overwrites), so if a later scene re-touched the same record, reapplying `before` from this scene's row is unsound — this only works one absorb-hop back.

## Implementation notes

- Store convention: scenes are markdown+frontmatter (`store/frontmatter.py`); JSON sidecars for chronicle/plot/relationships/changes.
- Backend tests: `backend/tests/test_scene_store.py` pattern, `monkeypatch.setenv("GRIMOIRE_HOME", tmp_path)`.
- Invariant: never touch `relationships.json` silently — no scene attribution exists there, so any "reversal" claim must be scoped to chronicle/plot/changes only.
- `_safe_id`/`SceneNotFound` guards in `scenes.py` should be reused, not reimplemented.

## Dependencies & related

- #78 — retcon-with-re-extraction reuses the same "which continuity records are scene-attributable" analysis.
- #79 — replay mode builds directly on a working cascade-delete-and-reconcile primitive.
- #72 — campaign fork is the safer alternative to destructive cascade delete for large scenes.

charlesmsiegel:
Closes #75.

A player can cut a scene at any post — that post and everything after it — and what the scene wrote to the campaign is taken back with it.

**`DELETE /campaigns/{cid}/scenes/{sid}/messages/{index}`**, mirroring `PUT .../messages/{index}`. 400 for an index that would remove nothing, 404 for an unknown scene. The reply is a **report**, not an `{"ok": true}` — see below.

**`store/cascade.py`** composes the whole operation under one campaign lock:

| step | what it does |
| --- | --- |
| `scenes.delete_from` | slices the transcript, clamps `turn_sizes` against *tracked* blocks, rewinds the scene's setting and clock |
| `commits.retire_scene` | fences a review prepared from the pre-cut transcript (runs *before* the cut) |
| `turnstate.supersede` | retires the per-post state ledger from the cut on |
| `alternates` | drops the parked reroll variants — unless the cut never reached the generation they are takes of |
| journal pass | reverses the scene's journalled write-backs, newest first |
| store sweep | chronicle record, plot + commitment beats, changes deltas, provenance citations |
| `scenes.unmark_absorbed` | the scene is unfinished again and can be re-absorbed |

## On the reversal — a note on the issue's own analysis

The issue sketched Option B (reapply `changes.json`'s stored `before`) and correctly rejected it: that log keeps only the **last** write-back per record, so once a later scene touches the same record its `before` is a lie.

The **change journal** (#31) is what that analysis was missing. It holds one row per write-back with the record's prior value *and* the value read back after the write, so the reversal puts the actual value back and its compare-and-swap **refuses** — rather than guessing — when the record has moved since. That makes Option B sound, so this implements it, with Option A's structural sweep behind it for the writes the journal's retention no longer reaches.

Because journal rows carry a scene id, a **relationship feeling or bond** written by this scene *is* attributable and is reversed. The issue's invariant still holds: nothing guesses at `relationships.json`, which has no scene attribution of its own — only rows that name the scene are touched.

## The scene's own setting and clock

`location_history` and `time_history` live in the scene's frontmatter, and the last entry of each is where and when the scene *is*. `chronicle.scene_facts` feeds them to the absorb prompt and the rolling-summary facts digest, and the context builder puts the location in front of the model — so a scene cut back past a move that stayed at the wharf would be prompted somewhere its transcript never goes.

Both are rewound, and only when the surviving transition lines account **exactly** for the history they would explain (n entries imply n−1 lines, since the first move is silent). `serialize.py` owns the two line formats so the writer and the reader cannot drift, and a scene whose lines do not tally keeps what it has — `turns._tracked_suffix_fits`' rule: validate first, let data that does not add up authorize nothing.

## What is deliberately left standing

`store/cascade.py`'s docstring accounts for **every** store `scene_refs.repoint` fans out to; a store that persists a scene id and is not named there is one nobody has decided about. The confirm dialog states the ones a player would otherwise assume:

- **`rolls.json`** — append-only by design. The cut removes a manual roll's transcript *line* (where `edit_message` refuses to rewrite one); the ledger entry survives.
- **`facts.json`** — the ledger models supersession, not deletion.
- **`timeline.md`** — append-only dated lines, no scene field to match on.
- **`prompt_log`** — an audit record of what was sent, still true.
- **`appearances`** — the cast still was cast. A join *is* narrated, so a cut can remove the line announcing one; reversing it means unlocking a version other scenes may be reading, which is well past what cutting a post asks for.
- **`audit` baselines** — the mechanism `undo` defers a sheet edit to; dropping it would make that edit irreversible.
- **`scene_ideas`** — a saved idea still became this scene.
- **A record the scene *created*** (`new_character`/`new_location`/`new_lore`). `store/undo.py` declines these and points at #75 for the cascade — and the cascade declines them too: a created character has since been cast, given a dossier and an appearance record, and may be named by later scenes. It is **reported**, not deleted.

A cut anywhere in an absorbed scene un-absorbs the **whole** of it: there is no record of which post produced which claim, so re-running the absorb over what is left beats a wrong partial reversal.

## Nothing after the cut may raise

The cut is irreversible, and `chronicle`, `plot` and `commitments` each parse with a bare `json.loads` over a file the user owns — so one stray byte in `plot.json` was a 500 landing on top of an already-truncated transcript, leaving the chronicle record describing posts that had gone. Every step past the cut is guarded and names itself in `failed`; the route reports it and the UI says so. Same posture `absorb.apply_edits` takes with the display logs it writes after its edits land.

## Frontend

A 🗑 in the post gutter beside `✎` — offered on dice-roll lines too, where Edit is not. The confirm counts the posts, names what a finished scene additionally loses, and names the roll log and timeline as surviving. The cut takes the same latch reroll and the roll paths take, so a second one cannot land against indices the first is still moving. Two outcomes the shortened transcript cannot show are reported as separate clauses: records that could not be **put back** (named, without claiming which of the two reasons applies), and cleanup that could not **run**.

## Tests

- `backend/tests/test_cascade_store.py` — 35 tests: the cut (indices, roll lines, `turn_sizes`, turnstate, alternates both ways, commit epoch), the setting/clock rewind (including a history the transcript cannot account for, and an actor-join line not mistaken for a move), the reversal (value put back, refusal reported not overwritten, no self-undo, a second cut is a no-op, another scene's records untouched, `last_scene` fallback including a status-only movement, the sweep covering what the journal cannot, created records), and the guards (unparseable JSON in one store, a garbled journal, a bad index costing nothing).
- `backend/tests/test_routes.py` — 4 route tests including the absorbed end-to-end path.
- `frontend/src/routes/CampaignView.test.tsx` — 6 tests (cut, declined confirm, confirm copy, refusal banner, in-flight latch, failed-cleanup reporting) plus an assertion on the roll-line test.

## Verification

**Rebased onto `main` (`5d31d65`)** — linear, three commits, no merge commit. `main` is moving quickly and this branch has been rebased onto it more than once; each time the only conflict has been `store/__init__.py`, where the facade's import list gains a module from each side (`cascade` here, and `search`/`semsearch`/`embed_space`/`usage`/`clock`/`birthdays`/`scenario` there). Both sides kept every time, and the contribution is byte-for-byte what it was before each rebase: **+1492/−6 across the same 20 files**.

Gate re-run in full against this base:

- `ruff check .` — clean
- `scripts/verify_templates.py` — 85/85
- backend `pytest` — 5344 passed. The one local failure, `test_atomic.py::test_a_read_only_record_is_not_silently_replaced`, is environmental: this container runs as root, so `chmod 0o444` does not block the write. It passes in CI.
- frontend `tsc -b` clean, `vitest run` — 1431 passed across 78 files
- the API-baseline and import guards pass over the combined facade

Two earlier red runs were pre-existing flakes, diagnosed in [a comment below](https://github.com/charlesmsiegel/grimoire/pull/341#issuecomment-5299244768) — both reproduced on clean `main` at the time. The frontend one (`a failed stream keeps the override`) has since been fixed on `main` by `331e625` and has not recurred in any run against a base that includes it.

**Not run:** the `/codex:*` review gates from `CLAUDE.md` — no codex binary in this container, and the Codex bot reports the account is at its review usage limit. Worth running locally before merge.

https://claude.ai/code/session_01Sv9zNpwheudazhXjSrxGPc


charlesmsiegel:
Merged `main` in (`a2274b7`) — the base had advanced 10 commits, and the earlier CI failure came from code that only exists on that side.

**Both CI failures I've seen are pre-existing flakes, not this branch.** Evidence for each:

**1. `backend (py3.11)` — `test_search_store.py::test_the_query_parser_is_linear_in_the_word_count`**

A wall-clock assertion (`< 0.5s` to parse 20k words); it measured 0.64s. `search.py` and its test arrived on `main` after this branch was cut, and nothing in this diff touches them. Decisive: **`backend (py3.14)` ran the same test on the same merge in the same run and passed** — six of seven jobs were green. Locally the inner parse ranges 0.05s–4.25s depending on machine load.

**2. `frontend` — `CampaignView.test.tsx > a failed stream keeps the override, and retry carries it`**

This one showed up locally after the merge and looked like mine, so I chased it properly:

- The test and `CampaignView.tsx`'s override path are untouched by this branch, and untouched by `main`'s new commits.
- The test mounts a scene with `messages: []`, so the control this PR adds to the post gutter never renders in it.
- Full-suite runs, my branch vs a clean `origin/main` worktree, interleaved to control for machine load: **`main` fails it too** — 1 in 13 runs on main, ~4 in 21 on this branch. Same ballpark, both non-zero.

So it's a load-sensitive race in that test (it races a `waitFor` against a `mockRejectedValueOnce` stream), and it predates this PR. I'm deliberately not "fixing" either test here — changing another feature's perf threshold or stream timing inside a cascade-delete PR would be scope creep, and both are worth their own look.

One caveat on my own first pass, since it's a trap worth naming: running a single test with `vitest -t` made it fail **deterministically** on both trees. That's an isolation artifact of the filter, not a real result — the whole file passes 322/322. Only full-file or full-suite runs mean anything for this one.

Verified locally on the merged tree: `ruff` clean, `verify_templates.py` 81/81, backend `pytest` 4993 passed, `tsc -b` clean, frontend 1366 passed. Integration with `main`'s new `store.search` was the one real merge point — `store/__init__.py` and `store_api_baseline.json` now carry both `cascade` and `search`, and the API-baseline and import guards pass.

---

PR #344, opened 2026-08-14
Auto-backup: the store zipped, kept, and swept (#32)

### LINKED ISSUE #32: Auto-backup (zip data + retention) and retention sweeps
Part of the **Storage & Configuration** milestone. From the pre-rebuild target feature list:

> Auto-backup (zip data + retention) and retention sweeps

## Current state

There is no backup mechanism anywhere. The entire app state is plain files under one root resolved by `backend/src/grimoire/store/paths.py:home()` (`GRIMOIRE_HOME` env → bootstrap pointer `~/.grimoire.json` → `~/.grimoire`), so a zip of that tree is a complete, restorable backup. The Configuration page already manages this root ("Storage location": `GET/PUT /api/config/data-dir` → `paths.data_dir_info()/set_data_dir()`, UI in `frontend/src/routes/ConfigView.tsx`).

Config lives at `<home>/config.md` via `store/config.py` — new keys must be added to `_CONFIG_KEYS` or `write_config()` silently drops them. There is no scheduler or background machinery: `main.py:_lifespan` only runs `migrations.migrate_scene_ids()`. A derived-data convention already exists at `home()/.cache/` (`store/thumbs.py:_cache_dir()` = `.cache/thumbs`) — rebuildable, excludable.

### Option A — opportunistic in-process sweeps (recommended)
- New `store/backups.py`: `create_backup() -> Path` zips `home()` (stdlib `zipfile`, deflate) into `home()/backups/grimoire-<UTC-timestamp>.zip`, excluding `backups/` and `.cache/`; `sweep()` deletes oldest archives beyond `backup_keep`.
- Config keys `backup_enabled`, `backup_interval_hours`, `backup_keep` (string values, like existing keys).
- Trigger: an asyncio task started in `main.py:_lifespan` that hourly checks "is the newest archive older than the interval?", plus one check at startup. No cron; works for however long the server runs.
- Routes: `GET /api/backups` (name/size/date list), `POST /api/backups` (back up now); ConfigView gains a Backups block beside Storage location, with typed methods in `frontend/src/api/client.ts`.
- Tradeoff: backups only happen while the app runs — acceptable since the store is only mutated while it runs.

### Option B — event-triggered safety backups only
- No schedule: back up before risky operations — data-dir change, world/campaign delete, sync `accept`, future retcon/bulk writes.
- Cheapest, catches the moments that matter; no steady-state protection. Composes with A as a later hook.

### Option C — tiered (GFS) retention
- Option A plus N-daily / M-weekly / K-monthly retention resolved from timestamped filenames during the sweep.
- Better long-horizon coverage per GB; more policy logic to test. Overkill for v1.

## Implementation notes

- The zip must exclude its own output dir (`backups/`) or archives recursively swallow prior archives; exclude `.cache/` too (thumbnails, any future index — see a general-purpose SQLite-derived index/cache substrate (not re-filed as its own issue)).
- The bootstrap pointer `~/.grimoire.json` deliberately lives *outside* `home()`; don't chase it — a restored tree works under any pointer.
- The store may live in a synced folder (explicitly supported per `CLAUDE.md`); `backups/` inside it multiplies sync traffic — consider an optional `backup_dir` config for an out-of-store location.
- Restore stays manual for v1 (unzip to a fresh dir, point Storage location at it via the existing UI); in-place API restore while the server is serving is race-prone.
- Timestamps: `paths.now_iso()` format contains `:` — use a filesystem-safe variant in filenames.
- Tests: `monkeypatch.setenv("GRIMOIRE_HOME", tmp_path)`, build a small store, assert archive membership/exclusions and sweep behavior with injected times (`backend/tests/`, pytest).

## Dependencies & related

- #31 — fine-grained undo journal; zips are the coarse layer beneath it.
- a general-purpose SQLite-derived index/cache substrate (not re-filed as its own issue) — keep the derived index under `.cache/` so it stays excluded automatically.
- #80 / #107 — "checkpoint before risky operation" flows could reuse Option B's pre-op hook.

charlesmsiegel:
Closes #32. Option A from the issue — opportunistic in-process sweeps, no cron.

The whole app state is plain files under one root, so a zip of that root *is* a complete restore point: nothing to parse, nothing else running.

## What's here

**`store/backups.py`** — `create_backup()` walks `home()` into `grimoire-<stamp>.zip`, excluding the backup directory (or archives swallow archives) and the rebuildable `.cache/`. `sweep()` deletes the oldest past `backup_keep`. `due()`/`run_scheduled()` are the schedule.

**Config keys** `backup_enabled` / `backup_interval_hours` / `backup_keep` / `backup_dir`, with accessors beside the existing ones.

**`main._backup_ticker`** — checks at startup and hourly, off the event loop, inside a task group cancelled on shutdown. No cron: the store is only mutated while the app runs, so "for however long the server is up" covers everything that can change.

**Routes** `GET /api/backups` (directory + listing) and `POST /api/backups` (back up now, then sweep — the response is the refreshed listing).

**UI** — a Backups section in Configuration's *The install* group: the four settings on the page's normal draft/Save cycle, plus a `BackupsPanel` listing the archives with a *Back up now* button.

## Decisions worth reviewing

- **Off by default.** An archive is a second copy of the library and the retention count multiplies it; in a synced folder each one is uploaded whole. That isn't a cost to impose behind an existing install's back, so the page asks — and says what it costs. Happy to flip this if you'd rather it default on.
- **Not a transactional snapshot.** Holding every campaign's lock for the length of a zip would stall play, so a backup taken mid-turn can catch one campaign's files a moment apart. Coarse-but-always-available beats consistent-but-blocking; the fine layer is #31.
- **`atomic.streaming_write`** is new: `write_bytes` takes finished bytes and a library doesn't fit in memory. It yields the temp's file *object*, never its pathname — the property that module was rewritten to have. `test_no_api_hands_out_the_temp_pathname` pins the module's public API as an exact list, so adding to it is a reviewed act; `append_line` (from main) and `streaming_write` now sit on it with a note saying why neither reopens the swap window, and a new test asserts the handle-not-path property directly.
- **The store root has its own lock domain now** (`locks.backup_lock()`, a leaf like `config_lock`) so two processes on one synced store don't zip it twice, and a sweep can't count a listing another run is publishing into.
- **`store_api_baseline.json`** gains `backups` — a deliberate facade addition, in the change that makes it.
- **The sweep runs after the archive, not before.** On a nearly-full disk that is the difference between a failed backup and a deleted one.

Restore stays manual for v1 — unzip into an empty folder, point Storage location at it. In-place restore while the server is serving is race-prone.

## Five rounds of adversarial self-review

The first pass shipped seven real defects. Each was found by attacking the diff and then **confirmed with a probe script before being fixed** — a review finding nobody reproduced is a guess. Commits `0dc8134`, `a4b7cb6`, `0e8c3a2`, `68f09ef`.

Backups that silently stopped or silently lied:

- **A directory grimoire couldn't read was dropped from the archive and the backup reported success.** `os.walk` swallows listing errors by default. The module docstring claimed in bold that this could not happen — true of files, false of directories, which is the bigger unit. A reviewer would have believed the comment.
- **A file that vanished mid-walk killed the whole backup.** Every atomic write in the store drops a temp and renames it away, so a backup taken while anyone was *playing* would fail. "Propagate everything" was right for an unreadable file and wrong for a vanished one.
- **An archive stamped in the future stopped the schedule forever** — a store synced from a machine whose clock is ahead, with nothing anywhere saying so.
- **A file named `grimoire-20269999T999999Z.zip` did the same, plus a bare 500.** The pattern accepted digit runs that aren't dates and the parse was a separate step that raised `ValueError`. Parsing is the recognition now.
- **One pre-1980 mtime failed the whole backup**, with a `ValueError` the route didn't catch. `strict_timestamps=False` clamps both ends — including the post-2107 case my first hand-rolled fix missed.

Wrong rather than broken:

- **A failed sweep was reported as a failed backup**, about an archive sitting right there, and threw away the listing that would have shown it.
- **With `backup_dir` set to the store root, the archive contained its own in-progress temp.**
- The backup directory was **resolved four times inside one `create_backup`** — four config reads assumed to agree — and `sweep` listed from one resolution and unlinked from another.
- `_norm` (a `getcwd` per call) ran twice per file across a whole library to re-answer a question the directory pruning had already settled.
- The panel's read-only text box was a focusable control nobody could type into, announced as an editable textbox. It's prose now.

And the gap none of that would have caught: **every test asserted archive membership, and membership is not the promise.** There's now an acceptance test that builds a real store through the store API, backs it up, extracts it elsewhere, points `GRIMOIRE_HOME` at the result and reads the library back.

## Testing

- `backend/tests/test_backups_store.py` (45) and `test_backups_routes.py` (15), including a regression test per defect above — each verified to fail without its fix.
- `test_atomic.py` — two new tests for `streaming_write`.
- `BackupsPanel.test.tsx` (10) and five new `ConfigView.test.tsx` cases.

Rebased onto `5d31d65`: `5366 passed` backend, `1439 passed` frontend, ruff and tsc clean. One pre-existing failure, `test_a_read_only_record_is_not_silently_replaced`, fails on `main` too in my container — it asserts `os.access` refuses a `0444` file, which root bypasses.

The branch is a linear rebase rather than a merge. The conflicts were real but additive — both sides adding config keys, Configuration sections, `store/__init__` imports, and (in `atomic.py`) a new public writer each. The one non-additive resolution is the `test_no_api_hands_out_the_temp_pathname` pin, which is a single sorted list and so had to be merged by hand rather than by keeping both sides. `d2e51b6` repairs a brace a union resolution ate, caught by tsc.

**Gates not run:** the repo's `/codex:*` review gates aren't available in this environment (no `.claude/commands/codex`), so the spec/plan/diff Codex checkpoints in CLAUDE.md were skipped. The five rounds above are not a substitute for them.

https://claude.ai/code/session_01BUKf7BKLwp6wq8Caoch1tC


charlesmsiegel:
CI red on `pydantic 1.10 (the Android dependency set)`, and it isn't this diff.

The failure is `test_search_store.py::test_the_query_parser_is_linear_in_the_word_count`, a wall-clock assertion that came in when main advanced (#332) — it measured 0.5617s against a 0.5s budget. `backend (py3.11)` runs that same test on the same commit and passed; only the pydantic-1.10 job, which builds a throwaway venv and runs the whole suite alongside it, was loaded enough to trip it. This branch touches nothing under `search`.

I've merged current main in and pushed, so CI is re-running against the fixed base. Locally the merged tree is `5015 passed` including that test.

Worth noting for its own sake: a 12% margin on a wall-clock budget will keep doing this on shared runners. The assertion is guarding against a quadratic parser, so it would still catch a real regression at a much looser bound — but that's #332's call, not something I'll change from here.

Also, since it's easy to miss in the diff: this branch has had five rounds of adversarial self-review since the PR was opened, and they found seven real defects in my own first pass — including two that made backups stop silently and one where the module docstring's central claim was false for directories. The commit messages carry the details.

---

PR #389, opened 2026-08-20
Re-establish the CI quality gates, blocking from day one, against a ratchet (#1)

charlesmsiegel:
Re-establishes the seven "clean carry-over" gates from #1 against the attempt-2
tree. Each one is **blocking from this commit** — none of them lands
report-only.

## The problem with the issue's own recipe

#1 proposes "land report-only, clear the surfaced set, then promote to
blocking". Re-derived against this tree, the surfaced sets are:

| Gate | Findings |
|---|---|
| ruff, selection widened from 5 families to 33 | 1183 |
| mypy, at its defaults, over the shipped package | 246 |
| typescript-eslint, type-checked rules | 949 |

That is a program of work, not a PR — and step one of the recipe is exactly
what `ci.yml` already argues against twice in its own comments: a job that
cannot fail is a job nobody reads, and the number only goes up.

## What this does instead

`scripts/ratchet.py` runs each tool and compares its findings to a committed
count **per (file, rule)** under `lint-baselines/`. New code is held to the
full rule set immediately — a file with no baseline entry must be clean — while
the existing backlog is recorded rather than suppressed.

It fails in both directions. An improvement leaves the recorded count stale, so
`make baseline` and the smaller file land in the same commit. And `--update`
**refuses to write a baseline that permits more** than the current one: that
direction needs `ACCEPT=1` typed out, so regenerating cannot quietly become the
way a red gate goes green. Deleting the file is not a way round it either —
absent counts as zero-of-everything.

Keyed by (file, rule) so the module carrying 87 `B904`s still fails on its
first new `F401`; not by line, or every insertion above a finding would read as
new.

## The gates

- **`check-lint`** — `select` goes from `F/BLE/E9/E402/RUF100` to 33 families.
  The ones deliberately left out are listed in `ruff.toml` with their count and
  the reason (`S`: 11633, nearly all `assert` in tests; `PLC`: 789 of 790 are a
  rule `test_import_guard.py` already enforces more precisely; `TID252` would
  fight the relative-import rule CLAUDE.md requires; `TC` would move
  annotations FastAPI resolves at runtime). C901 is on at mccabe's default 10,
  which is also this tree's exact p99 — 89 of 9142 functions are above it, and
  they are itemized in the baseline rather than argued away with a larger
  number. B008's 35 findings are 29 FastAPI dependency markers, which are
  declared immutable rather than suppressed one by one.
- **`check-mypy`** (new CI job) — the shipped package, at mypy's defaults.
  `mypy.ini` names the four untyped third-party packages individually rather
  than setting `ignore_missing_imports` globally, since a global one also
  silences a first-party module that fails to import. The two that live in
  optional extras are `follow_imports = skip`, so the count cannot depend on
  whether the reader installed `[desktop]`. Sampling the findings: they are
  largely invariants the code states in a comment and mypy cannot prove ("both
  are always set"), not live crashes — but that is a sample, not an audit,
  which is what the baseline is a work list for.
- **`check-eslint`** (new CI job) — there was no eslint at all. It arrives with
  typescript-eslint's **type-checked** rules (the syntax-only ones would find
  nothing; tsconfig is already `strict`), **jsx-a11y** recommended,
  **`react/jsx-key`** and **`react/no-array-index-key`** for the stable-key
  pair, and **react-hooks**, which 31 `eslint-disable` comments in the tree
  were already written for. The `any`-shaped rules are off in test files and in
  `src/testkit/`: a vitest module mock *is* `any`, and leaving them on would
  have blocked the next test file for using the idiom every existing one uses.
- **`check-py`** — now runs under pytest-cov with `--cov-fail-under=93` (the
  integer below the 93.11% the suite reaches), and CI uploads
  `backend/coverage.xml` as `backend-coverage`.

## What was cleared rather than recorded

- ~520 ruff findings, by `ruff check --fix` — in their own commits, so the
  import rewrap does not bury the rest of the diff.
- 18 frontend findings whose right answer is unambiguous: three `alt`
  attributes that said "image", a `<label>` captioning a radio group it owns no
  control in, two vitest mocks whose hooks `rules-of-hooks` could not see, four
  `eslint-disable` directives that no longer suppress anything, and three
  unused bindings.

`react/jsx-key` is at zero and stays there — it is the one rule here with no
baseline entries at all.

## Eight rounds of adversarial self-review

The Codex gates could not run (usage limit — see the bot's comment), so the
diff went through eight review rounds instead, each landed as its own commit.
Four found real defects in the harness this PR is *about*:

1. `make baseline` rewrote the baseline from whatever the tree said, so the
   ratchet was a ratchet in one direction and a free pass in the other. Also:
   nothing checked which ruff or mypy produced the counts, `json.loads` ran
   straight off the tool's stdout (an `npx` diagnostic became a
   JSONDecodeError), and no test touched any of the three parsers — there are
   now 14.
2. The eslint gate would have rejected the next test file for using the mock
   idiom the others use. 7710 of 8538 findings were in tests; scoping the
   `any` rules honestly took the baseline under 1000.
3. The version check asked whether the pin appeared *anywhere* in the output,
   so ruff 10.16.0 satisfied a pin of 0.16.0.
4. `check-eslint`'s own `npm ci` — with a comment claiming that made it safe
   under `make -j`, when it was the collision. Plus CRLF baselines on Windows
   and a missing baseline being treated as an empty one.

Rounds five to eight: the same `-j` hazard in `web-dist`; `grimoire` being
sorted as a third-party package inside its own repository; a cancelled CI run
reporting a red X for a coverage report it never got to write (caught by CI, on
this PR); and `make check` making you wait eighteen minutes to hear that a
nine-second gate is red.

## Rebased three times, and what that taught

Main landed 55 commits while this branch ran, so it was rebased onto caf8e64,
then 5118305, then cfbb899. Two things worth knowing:

- **CI checks out the *merge* of a PR into main, not its head.** So the
  baselines have to describe main-as-of-now, and an earlier red run here was
  exactly that — CI seeing main commits the branch had not rebased onto yet,
  reported as findings that had *disappeared*. That churn is a property of
  landing a gate main does not have: once these are on main, a commit cannot
  move a count without updating the baseline in the same commit, because the
  gate is what lets it merge.
- **`src/testkit/` was a real gap.** Main extracted the shared `vi.mock`
  factories into a directory that is not `*.test.tsx`, and 186 findings came
  in with it — every one the mock idiom. Round two's argument, one directory
  over.

The rebases also found one more harness defect: `make baseline` did not take
`frontend-deps`, so it would regenerate the eslint baseline against whatever
`node_modules` happened to be installed — writing counts the gate cannot
reproduce, which is the only thing the file is for.

## Verification

- `make check-lint`, `check-mypy`, `check-eslint`, `check-templates` — pass
  locally and in CI, which is the reproducibility claim the baselines rest on.
- `make check-web` — 99 files, 1833 tests, green.
- `make check-py` — 6329 passed; `Required test coverage of 93% reached. Total
  coverage: 93.11%`.
- `make check-pydantic1` — venv built, `pip check` clean. Run because the `dev`
  extra gained two packages and that target resolves it from scratch under
  pydantic 1.10.
- `make -j4 check-web check-eslint` runs `npm ci` once; it ran it twice before.
- Both ratchet directions and both `--update` refusals were exercised
  end-to-end, and `test_ratchet_guard.py` covers them plus the parsers and the
  baseline files' shape.
- One pre-existing failure throughout,
  `test_a_read_only_record_is_not_silently_replaced`, is an artifact of this
  container running as root (root ignores the `0o444` the test sets); it fails
  identically on `main` with none of these changes applied.

## Not in this PR

The five items #1 files under "revisit once attempt-2 has the matching layers"
(#525, #526, #527, #528, #598) — they are waiting on layers this tree does not
have yet, and nothing here changes that. #1 stays open for them.


charlesmsiegel:
Two things worth saying rather than passing over.

**The Codex gate could not run.** `/codex:review` and `/codex:adversarial-review` both returned a usage-limit message on this branch (the bot's comment above). CONTRIBUTING says an unrunnable gate is a thing to state in the PR, so: stated. In its place the diff went through six adversarial self-review rounds, each landed as its own commit with what it found — the first four were real defects in the harness itself, including `make baseline` being an unconditional "accept everything" button and a `make -j` hazard whose Makefile comment claimed to be the fix for it.

**`pydantic 1.10` went red on 81ed548, and it is not this PR's.** The failure is `test_the_query_parser_is_linear_in_the_word_count`, which asserts a wall-clock budget:

```
assert time.perf_counter() - started < 0.5
E   AssertionError: assert (618.717554585 - 617.957844308) < 0.5
```

0.76s against a 0.5s budget on a shared runner. `search.query_terms` is untouched by this diff — it still has the `seen` set the test is about — and the only change to `test_search_store.py` is the import rewrap. The same job passed on this PR's two earlier commits and passes locally.

The test is guarding against a quadratic parse that took **six seconds** on a 60k-word paste, so its budget could be 3s and still catch everything it was written to catch; at 0.5s it is a wall-clock assertion competing with the rest of the suite for a 2-core runner. Proposed patch, deliberately not applied here since it is unrelated to this change:

```python
    # 0.5s was tight enough to fail on a loaded CI runner at 0.76s. The
    # regression this guards took six seconds, so a budget an order of
    # magnitude above the linear cost still catches it and stops measuring
    # the runner's load.
    assert time.perf_counter() - started < 3.0
```

Every other job is green on that commit, including `backend (py3.14)` — which is the one that had to clear the new coverage floor on the other interpreter.

---

Issue #235, opened 2026-07-28
Absorb is a long non-atomic, non-idempotent multi-phase write


charlesmsiegel:
Surfaced by the codebase atlas (`docs/codemap.html`, generated at 81b61ff1). **Ranked risk #3.**

## Problem

`POST /campaigns/{cid}/scenes/{sid}/absorb` is a multi-phase write with no transaction, no rollback, and no idempotency guard:

1. `store.absorb.materialize(cid, sid, parsed)` at `backend/src/grimoire/routes.py:3132` commits chronicle, character-state, relationship, plot and group edits **to disk**.
2. *Then* the per-NPC dossier loop runs — one LLM call per present NPC (`routes.py:3135-3144`).
3. *Then* the mechanics audit runs (`routes.py:3147`).

## Consequence

If the process dies (or the user closes the tab, or an upstream call hangs) during step 2, the campaign has absorbed the scene — chronicle updated, state advanced, plot threads moved — but some NPCs still carry pre-scene dossiers. Re-running absorb re-applies every step-1 edit from scratch on top of the already-absorbed state.

There is no "already absorbed" check on the route, and no idempotency key. The chronicle is keyed by scene id so it overwrites cleanly, but the state/relationship/plot edits are applied as deltas.

## Suggested fix

Minimum: guard the route against re-absorbing a scene that already has a chronicle entry, and return the prior result (or require an explicit `force` to re-run). That alone converts the failure from silent corruption into a no-op.

Better: stage the step-1 edits and commit them only after the dossier loop and audit complete, so an interrupted absorb leaves the campaign untouched.

Related: the dossier loop's silent failure handling is tracked separately.

Issue #271, opened 2026-07-29
PUT /chronicle is a non-atomic, replayable commit


charlesmsiegel:
Split out of #235, which fixed the *proposal* side (`POST /absorb` is now
side-effect-free and guarded against re-absorbing). The commit side still has
the failure class named in that issue's title.

## Problem

`PUT /campaigns/{cid}/scenes/{sid}/chronicle` (`routes.py`, `put_chronicle`)
performs four sequential writes with no transaction and no rollback:

1. `chronicle.absorb(...)` — the chronicle entry
2. `chronicle.append_timeline(...)` — appends
3. `scenes.mark_absorbed(...)` — scene frontmatter
4. `absorb.apply_edits(...)` — every approved edit, then `changes.record(...)`

Each individual write is atomic (`store.atomic`), but the sequence is not.

## Consequences

- **Crash mid-sequence** leaves an "absorbed" chronicle entry with the state,
  plot and dossier edits unapplied. #235's new 409 guard then refuses an
  ordinary re-absorb of that half-committed scene (recoverable only with
  `?force=true`).
- **A replayed PUT is not idempotent.** `append_timeline` appends and
  `plot.set_movement` adds a beat per apply, so a second commit of the same
  review duplicates both. #235 added a `saving` latch in `CampaignView`, which
  stops a double-click — it cannot stop a retry after a lost response, a
  second tab, or a direct API caller.
- **Non-sheet apply failures are swallowed** (`apply_edits`' per-edit
  `except`). Only sheet edits have a failure contract, so a failed dossier or
  plot write returns 200 with no indication anything was skipped.

## Suggested fix

A one-shot commit token: `POST /absorb` returns one, `PUT /chronicle` requires
and consumes it, and a replay returns the prior result instead of re-applying.
That closes the replay half. Crash-mid-sequence additionally needs the four
writes staged and published together (or a small journal).

Worth pairing with a failure contract for non-sheet edits so a partial apply is
reported rather than silently returning 200.

charlesmsiegel:
Sharpening the third bullet, because #235 moved a guarantee into this gap.

#236 gave the dossier refresh an explicit failure contract: an exception from
`dossiers.write()` landed in `dossiers.failed` and drove the phase to
`degraded`/`failed`, which the scene inspector renders.

#235 moves the dossier *write* out of `POST /absorb` and into
`apply_edits()` at commit time. Generation failures (rate limit, parse error,
[...]

Issue #95, opened 2026-07-11
Turn cancel / timeout / heartbeat + transactional post+response


charlesmsiegel:
Part of the **Scenes & Play** milestone. From the pre-rebuild target feature list:

> Turn cancel / timeout / heartbeat + transactional post+response

## Current state

None of cancel, heartbeat, or transactional post+response exist; the turn loop has a real data-loss gap.

- **No cancel.** `_chat_stream`/`_ephemeral_stream` (`backend/src/grimoire/routes.py`) wrap `OpenRouterClient.stream()` in a plain `async def event_stream()` generator with no `asyncio.CancelledError`/`GeneratorExit` handling. If the client disconnects mid-stream, Starlette closes the generator (raising `GeneratorExit` at the current `yield`), which is **not** `OpenRouterError` — so the `except OpenRouterError` branch (`routes.py:1327-1330`, which *does* persist partial `parts` on that specific error) never runs, and the partial reply is silently dropped. There is no client-side abort affordance either (`frontend/src/api/stream.ts`'s `streamPost` has no `AbortController`/`signal` plumbing).
- **No timeout enforcement beyond the HTTP client's own.** `openrouter.py:OpenRouterClient._client()` sets `httpx.Timeout(120.0, connect=30.0)` — a hard-coded 2-minute read timeout with no per-call override, no user-visible "this is taking a while," and a timeout surfaces only as a generic `network`-kind `OpenRouterError` after the fact.
- **No heartbeat.** SSE deltas are the only events; a slow-starting generation (before the first delta arrives) looks identical to a hung connection to any proxy/load balancer with its own idle timeout, and the client has no way to distinguish "still working" from "stalled."
- **User post precedes the stream, not the response.** `post_chat` (`routes.py:1452-1469`) appends the user's message (`scenes.append_message(..., "user", ...)`) **before** calling `_chat_stream` — confirmed by reading the route body directly. If the subsequent stream then fails in a way that isn't a clean `OpenRouterError` (the `GeneratorExit`/disconnect case above, or any other unhandled exception), the user's post is left in the transcript with no reply and no error surfaced — not transactional. The empty-content/pcless "director turn" path (`build_director_messages`) is the one case that's already safe here, since its note is explicitly never persisted.

### Option A — layered fix: exception-safe persistence + cancel signal + heartbeat (recommended)
[...]

PR #393, opened 2026-08-20
Spec: detached runs, so every LLM call survives a locked phone

charlesmsiegel:
**The design spec only — no code.** One file, 1374 lines:
`docs/superpowers/specs/2026-08-20-detached-turn-runs-design.md`.

The spec lands on its own. The plan and implementation follow separately, so
this PR is reviewed for whether the design is right, not for whether code
matches it.

## The problem

An LLM job's lifetime is currently the HTTP request's lifetime. On Android
that means locking the phone or switching apps loses it. Two failure modes
that look identical to the player and are mechanically opposite:

- **A chat turn loses the work.** `_chat_stream` drives `client.stream(...)`
  from inside the `StreamingResponse` generator. The client goes away,
  Starlette cancels the task, `_flush_on_abort` keeps the partial, and
  generation stops. There is nothing still running to re-attach to.
- **Ending a scene loses the result.** `post_absorb` is a plain endpoint, so
  uvicorn runs it to completion — but it is `@computes_only`, so by the
  deliberate design of #235 the entire review lives in a response body nobody
  receives. Ten minutes of fan-out, discarded. `post_audit` and
  `post_dossiers` go further and actively abort on `request.is_disconnected`.

And nothing keeps the process alive: `ServerService` is a plain started
service and the manifest declares only `INTERNET`.

## What the spec proposes

**Every LLM call becomes a run** — all 25 LLM routes. Two orthogonal
properties carry the design:

- **Subject** — `scene(cid, sid)`, `campaign(cid)`, `world(wid)` or `global`.
  It cannot be `(cid, sid)`: taglines, voice anchors, image drafts and scenario
  parse are world-scoped, suggestions and intent are campaign-scoped, and the
  model refresh has no campaign at all. The subject decides which lock a run's
  persistence takes.
- **Class** — `turn`, `review`, `background`, `draft` — carrying exclusion,
  durability and notification policy so it isn't restated per route.

A run is **one route invocation**, not one provider call: absorb's fan-out
helpers execute inside the parent run and reserve nothing, or an absorb would
deadlock against its own exclusion key.

Runs start inside the lifespan task group — bridged from the six synchronous
streaming handlers, each behind its own failure boundary. Streaming runs tee
absolute-indexed frames into an append-only buffer, so a foregrounded client
tails live exactly as today and a returning one replays from its own consumed
index. Absorb results become durable in a new `store/pending_reviews.py`. On
Android, foreground-service promotion while any run is live plus a local
notification when a turn or review lands — no FCM, no cloud, no accounts, per
§9 of the architecture doc.

It also completes two things specified and left unbuilt:
`docs/android-architecture.md` §4 / risk 6 / Phase 3, and the
foreground-service item in the scope of
`2026-08-18-absorb-performance-design.md`.

## Review history

**Seven rounds — one local, six from Codex — with 52 findings, every one
verified against the code before being acted on, and all 52 resolved.**
Counts by round: 10, 11, 8, 8, 7, 3, 5.

The rounds that mattered found design holes, not wording:

- **The scheduling mechanism could not work as first written.** All six
  run-producing streaming handlers are synchronous `def`, so FastAPI runs them
  in a threadpool worker, and `start_soon` on an AnyIO task group is not
  thread-safe. Every run would have failed immediately.
- **Scene ids are recycled** — `serialize.py:_numbering` derives the next
  number from files on disk with no stored counter — so an existence check
  would let an old run publish onto a *different* scene.
- **The commit epoch does not move when play continues.** `commits.reserve`
  runs only from `PUT /chronicle`, so a landed review could be saved against a
  transcript that had grown past it.
- **One regression the spec had invented** — a `no module resolved` pre-flight
  on `post_absorb`, which has no such guard. It would have made End Scene
  impossible for every module-less campaign.

Rounds five through seven were overwhelmingly findings about fixes from
earlier rounds, which is where the review was stopped: what remains is
implementation-shaped and belongs in the plan.

## Known limits, stated in the spec

Persistence covers a *finished* absorb whose client went away. It does **not**
survive a process death at minute eight of a ten-minute fan-out — that would
need phase-level checkpointing and is explicitly out of scope. The foreground
service is the only defense against that case.

https://claude.ai/code/session_01AEyrMtX8M3dVrAV4yEZSzV


charlesmsiegel:
All ten findings verified against the code before changing anything, and all ten hold. Addressed in `d7c6066` and `76a713d`; resolving the threads individually.

Two of them found holes in fixes this spec had *added*, which is the useful kind:

**`Correlate recovery with the attempted POST` (P1)** — correct, and it broke my own #95 fix. "A run exists within the reap window" was offered as proof a lost response's post had landed. A run that finished eight minutes ago is still in the window, so a send that never reached the server would find it and report as accepted — an "I don't know" converted into a confident wrong answer, which is worse than the gap being closed. The POST now carries a client-generated attempt id recorded on the run, and recovery asks about *that id*. Fallback when it ages out is refetch-and-compare, the honest answer. This also demotes the ten-minute window from a correctness knob to a convenience one.

**`Reserve the run slot before every route mutation` (P1)** — correct. The spec had the check before `post_chat`'s append, which misses that retry heals and supersedes proposals, regenerate archives and removes a reply, replay stages posts, and roll-proposal resolves a check. Reserving late there is worse than not checking: the player is told nothing happened when something did. Now reserved before the first mutator in every run-producing route, released if the synchronous setup that follows fails.

The rest, briefly:

| Finding | Resolution |
|---|---|
| Resume from the subscriber's consumed index (P1) | Client tracks its own consumed index; `next_index` reports the live tail only. Resuming from it would have lost the middle of a reply in exactly the locked-phone flow this exists to fix. |
| Preserve the full review when retry runs land (P1) | Confirmed at `useSceneReview.ts:446` and `:521` — both fold partial payloads into the absorb. Retry persists are now a read-modify-write under `campaign_lock(cid)`, so prose, staged edits and `commit_token` survive. |
| Validate every run ID against the route's scene (P1) | Every lookup compares the run's stored `(cid, sid)` to the path and returns `run_gone` on mismatch. Addressing by id closed one direction; this closes the other. |
| Move pending reviews when scenes are renamed (P1) | Joins the `scene_refs.repoint` fan-out and is named in `store/cascade.py` — whose stated rule is that a store persisting a scene id and not named there "is one nobody has decided about". A harder obligation than advisory. |
| Serialize review cancellation with result persistence (P1) | Cancellation is recorded before the delete, and the terminal persist checks it — both under one `campaign_lock(cid)` hold, so a dismissed review cannot reappear. |
| Record campaign activity when a pending review is written (P2) | Terminal persist stamps activity itself; `@computes_only` stays correct for the 202, which really does return nothing. |
| Capture notification labels before the scene can disappear (P2) | Labels captured at run start, and the notification hook isolated with non-fatal failures — it must never interrupt the persist or foreground-service demotion. |
| Preserve the retry endpoints' preflight errors (P2) | Generalized from `post_absorb` to all three, naming each guard including the dossiers empty-transcript one that was missing. |

Each also gained a test in the spec's testing section, including one written to fail on the stale-index bug specifically: it asserts frame contiguity rather than "some frames arrived".

---

charlesmsiegel:
Round two: all eleven verified against the code first, all eleven hold. Addressed in `5ec23ac`.

**Three deserve calling out, because they were holes in round one's fixes or in my own reading:**

**`Contain runner failures inside the lifespan task group` (P1)** — correct, and it undercut the fix I'd just made. Starting runs in the lifespan group gives supervised shutdown, but an unhandled exception in an AnyIO child cancels every sibling and propagates out of `_lifespan` — so one malformed scene would abort every other live run, stop the backup ticker, and potentially end the process. Each runner now has its own boundary: catch everything that isn't shutdown cancellation, mark `failed`, do the normal terminal bookkeeping, log, return.

**`Keep module-less campaigns absorbable` (P1)** — correct, and this one was a regression I invented. I listed a 400 "no module resolved" as a `post_absorb` pre-flight. It has no such guard, deliberately: `_run_audit` returns a skipped mechanics block with reason `"no module"` (`scenes.py:1138`) while extraction, dossiers and voice complete normally. That check lives only in `post_audit`. Shipping it would have made End Scene impossible for every module-less campaign. Removed, with the reasoning written down so it doesn't get re-added.

**`Count skipped SSE frames in the reconnect cursor` (P1)** — correct, and it's the mirror of round one's frame-index finding. `parseSSEChunk` drops heartbeat comments without invoking a callback, so a cursor counted per `ChatEvent` lags the server's raw position by however many heartbeats went by — and a generation waiting on the model emits many. Resuming from that short index duplicates already-rendered text. Frames now carry an absolute index, so the two sides don't have to agree on what counts.

The rest:

| Finding | Resolution |
|---|---|
| Hold the run slot until cancellation cleanup finishes (P1) | State moves to `cancelled` only after `on_abort` returns; the cancel response doesn't resolve before that. Freeing at request time let a fast re-send race the partial-persist — two writers on one transcript. |
| Specify the retry merge at edit-row granularity (P1) | Confirmed at `useSceneReview.ts:452` and `:533`. Audit replaces only `sheet` rows; dossiers replaces only rows whose target is in `res.dossiers.proposed`, preserving prior proposals for NPCs it didn't re-propose; both move their `phases` row. Spelled out, since getting it wrong fails quietly. |
| Reject retries that have no pending review (P1) | Added to both retries' synchronous pre-flight. A 202 accepting work whose result the design has already promised to refuse is what the pre-flight rule exists to prevent. |
| Recheck scene existence before publishing a review (P1) | The runner rechecks under the same `campaign_lock(cid)` hold as the pending write. The cited guards protect transcript persistence; a sidecar write needed its own. |
| Preserve run_id on streaming 409 responses (P1) | `streamPost` reads the body then drops it, keeping `kind` and discarding `run_id`. Either it retains the payload or lifts `run_id` explicitly; the 409-to-attach path gets a test. |
| Bind review cancellation to the intended run (P1) | A review-generation id, minted at absorb start and stored on the review. "Most recent run" would have cancelled an unrelated chat run; "the run named by the record" finds nothing before the absorb publishes. Deletion stays idempotent. |
| Register notification channels before foreground promotion (P1) | Two channels with explicit ids, created at service start. An unregistered channel invalidates the foreground notification the process-lifetime guarantee rests on — not cosmetic. |
| Clear pending reviews on replayed successful saves (P2) | Deletion moved onto the `prior["done"]` idempotent path as well as the fresh-success tail, under the same lock. |

---

**Separately, the scope has widened at the author's request: every LLM call becomes a run, not just scene turns and absorb.**

An AST sweep found 35 functions holding an `LLMClient` across 25 routes, and most are not scene-scoped — taglines, voice anchors, image-description drafts and scenario parse are world-scoped; suggestions and intent are campaign-scoped; the model refresh has no campaign at all. So a run now carries a **subject** (scene / campaign / world / global), which decides what lock its persistence takes.

Policy moved to a **class** — `turn`, `review`, `background`, `draft` — and this made the document shorter rather than longer. The previous draft needed a prose taxonomy of which calls were runs and which were "deliberately not," including a warning that making the rolling summary a run would strand the composer after every turn. `background` declares no exclusion key, so that bug is now structurally impossible instead of documented, and the `_ephemeral_stream` carve-out disappears the same way.

One risk this creates, flagged in the spec: `test_usage_guard.py` recognizes an LLM call by the receiver name `client` inside `routes/`. Moving 35 call sites behind a uniform runner is precisely what that convention does not survive by accident, and a guard that fails open over the token ledger is worse than no guard. Either the meter stays at each call site or the guard is rewritten to follow runs — and it must be shown failing on a deliberately unmetered run before the work is done.

---

charlesmsiegel:
@codex review

Since the last round the scope widened substantially, so this is a review of a materially different design rather than a re-check of the previous one. Worth attacking:

- **Every LLM call is now a run** — all 25 routes, not just scene turns and absorb. A run carries a **subject** (`scene` / `campaign` / `world` / `global`), because most call sites are not scene-scoped: taglines, voice anchors, image-description drafts and scenario parse are world-scoped; scene suggestions and intent are campaign-scoped; the LLM-connection model refresh has no campaign at all. The subject decides which lock a run's persistence takes — and world and global subjects take none, which is new and worth probing.
- **Policy moved to a class** — `turn`, `review`, `background`, `draft` — with only `turn` and `review` declaring an exclusion key, which they share per scene. The claim to test is that `background` (rolling summary, scene break) declaring no key is what keeps those from stranding the composer after every turn.
- **The opener was reclassified** from `turn` to `draft` this round, on the grounds that `post_opener` persists nothing and `post_first_post` is what writes the accepted text.
- The eleven findings from last round are addressed in `5ec23ac`; the runner failure boundary, the absolute frame index, the edit-row-granular retry merge and the review-generation id are all new surface that has not been reviewed yet.

---

charlesmsiegel:
Round three: eight findings, all verified against the code first, all eight hold. Addressed in `adef1db`.

**Two turned on codebase facts rather than reasoning, and both checked out.**

**`Invalidate reviews when the transcript changes` (P1)** — the subtlest hazard in the design so far. `commits.reserve` is called from exactly one place, `PUT /chronicle` (`scenes.py:2629`), so the commit epoch advances **only on a chronicle-save claim** — appending a turn does not move it. Once a review lands its exclusion slot is released and the composer re-enables, so play continues while the review sits on disk, and its stale token still passes both the read-time check and the save. The scene gets marked absorbed while every post made after the review was prepared went unreviewed — silently, since every check involved returns green. The review now carries a transcript watermark validated under the campaign lock at retrieval and at save. This does not overlap with the epoch: #271 guards against a competing review committing; this guards against play continuing.

**`Verify scene identity rather than existence` (P1)** — correct, and the mechanism is worse than described. `scenes/serialize.py:_numbering` derives the next scene number from **the files on disk with no stored counter** (`top + 1`), so deleting the highest-numbered scene frees its number and the next create takes it back; with the same title slug that reproduces the identical `sid`. The existence check I added last round would pass against a *different* scene and publish the old review onto it, and the turn finalizer could append old narration to a replacement transcript the same way. Runs now capture an immutable scene identity at start and compare before any terminal write — scenes gain such a field, since number, title and position are all mutable by design (rename, `repad`, the front door's date slug).

**Six of the eight were created by widening the scope to every LLM call**, which is worth recording as a cost of that decision rather than a coincidence:

| Finding | Resolution |
|---|---|
| Track every concurrent run for a subject (P1) | The subject index was still a most-recent pointer — fine when only one exclusive run per scene existed, wrong once `draft`/`background` declare no key and legitimately overlap. Two image-description drafts in a world, and the first had no discovery path. It's a collection now. |
| Add collection routes for non-scene runs (P1) | Every run route required `cid` *and* `sid`, so campaign-, world- and global-scoped runs could not be discovered, polled, streamed or cancelled at all. I widened the subject and never widened the routes. The four operations are now mounted per subject level. |
| Propagate streamed error outcomes to run state (P1) | `_fence_stream` catches `LLMError`, emits an SSE `error` frame and returns normally. A runner inferring success from "did not raise" would mark failed turns `landed` and fire the success notification. The producer returns an explicit terminal outcome. |
| Make repeated attempt IDs idempotent (P1) | Recording the id made recovery queryable but not the POST idempotent — a duplicate delivery after the run went terminal finds the slot free and appends the prompt twice. Starting is get-or-return-existing on `(subject, attempt_id)` across the reap window. |
| Resume after the last consumed frame (P1) | Off-by-one: `from=N` replays frame `N`, so "resume from the last index seen" duplicates it. `from` is inclusive, the client persists `last_seen + 1`, and the contiguity test pins the boundary at every offset. |
| Serialize live runs with scene renames (P2) | Rename refused while the scene's exclusion key is held. Repointing a live run and its hooks mid-flight is easy to get subtly wrong; refusing for the seconds a run is live is not, and renames outside a live run — the ordinary case — are unaffected. |

Also closing the round-two threads that were addressed in `5ec23ac` but whose resolution calls didn't land at the time.

---

charlesmsiegel:
Round four: eight findings, verified against the code first, all eight hold. Addressed in `2c08793`.

**`Define how nested absorb calls share their parent run` (P1)** is the one that mattered — a real contradiction in "every LLM call is a run," and it needed settling rather than patching. `post_absorb` fans out to extraction, `_stage_dossiers`, `_stage_voice_drift` and `_run_audit`, and two of those issue a provider call *per NPC*. If each is a run it collides with its parent's scene exclusion key and an absorb deadlocks against itself; if none are, the "35 call sites" framing is false.

Resolved by fixing the unit: **a run is one route invocation.** The 25 routes are runs; the 10 helpers execute inside whichever run called them and reserve nothing. The widened scope means every LLM operation *a user initiates* survives a locked phone — never that the dossier call for the fourth NPC is independently resumable, which wouldn't help anyone: the absorb's budget and `_gather_phases` already own that layer, and "cancel one NPC's dossier" is not an operation that exists.

**Two were inconsistencies the scope change left behind in my own document.** The canonical run record still required `cid` and `sid`, still said every run belongs to exactly one campaign, and still listed only the original scene kinds — so a literal reading couldn't represent a world draft or the global model refresh at all. And the registry was described as living in a module without saying where the *instance* lives; module-global is the natural reading and is wrong here, since this repo builds and tears down `create_app()` repeatedly in one process and stale terminal records, attempt ids and exclusion keys would leak between apps.

| Finding | Resolution |
|---|---|
| Migrate existing scenes to immutable identities (P1) | Assigning at creation only would be **worse than nothing**: every existing scene predates the field, so an old scene and the replacement recycling its `sid` both present the same absent value — the check passes while reading as solved. Backfilled under the campaign lock, joining `_lifespan`'s existing idempotent migration pass, with a lazy backfill for campaigns that were busy. |
| Serialize automatic repadding with live runs (P1) | `repad` is a second, *automatic* rename path: `_create_scene` crossing a width boundary renames every scene in the campaign. Refusing the manual rename route never covered it. The width-changing create is refused while any exclusion key in that campaign is held. |
| Trigger post-turn background runs from the server (P1) | `CampaignView.askAfterPost` fires them once the streaming promise settles — precisely what does not happen when the phone locks. A scene break could then be missed indefinitely, since the next opportunity only comes with the next send. The turn runner schedules both after its own persistence. |
| Adapt every non-streaming draft caller to polling (P1) | Confirmed: `TaglinePrompt.tsx:17` destructures `{ tagline }`, `ConnectionEditor.tsx:106` reads `result.rev`/`models`/`fetched_at`, `SceneIdeaPicker` consumes `sceneIntent` inline. One shared start/poll/unwrap helper keeps each call site's existing shape and gains re-attachment; on mount each adopts an in-flight run rather than starting a second. |
| Preserve Retry-After across polled failures (P2) | Confirmed at `common.py:168`. Over a poll boundary a header describes the *poll*, not the failed call, so the delay is stored with the terminal error and returned in the payload. A rate limit is the one error where losing "try again in N seconds" changes what the player should do next. |

---

charlesmsiegel:
Round five: seven findings, verified against the code first, all seven hold. Addressed in `98c405a`.

**`Bridge synchronous route starts onto the lifespan event loop` (P1) is the most consequential finding of the whole review — it made the spec unbuildable as written.** Every run-producing streaming handler is a synchronous `def`: `post_chat` (`scenes.py:359`), `post_retry` (`:419`), `post_regenerate` (`:491`), `post_replay_turn` (`:3212`), `post_opener` (`greetings.py:318`), `post_roll_proposal` (`mechanics.py:111`). FastAPI runs those in a threadpool worker, and `start_soon` on an AnyIO task group is not safe to call from another thread. "Start runs inside the lifespan group" would have failed on every run, immediately. Either those six become `async def` or the start is bridged back onto the lifespan loop — the choice belongs in the plan, but the assumption that handler and task group share a thread had to go. The absorb family is unaffected; those routes are already async.

Two more that are about the real code rather than about the document:

- **`Buffer uploaded scenario cards before returning 202`** — `post_scenario_parse` reads an `UploadFile` (`worlds.py:616`). After a 202 FastAPI may close it while the runner is still queued, so a coroutine capturing it fails *after* acceptance — the client holding a run id for work that can never succeed. Now a rule: a run's coroutine owns its inputs and never captures request-scoped resources.
- **`Reserve the scene before capturing the absorb snapshot`** — snapshot-then-reserve leaves a gap where a fast chat turn can reserve, append, finish and release, after which the absorb is accepted against a transcript that has already moved. The watermark would refuse it, but only after the entire budget was spent. Reserving first prevents for free what the watermark would otherwise catch expensively.

**Four were holes in fixes from rounds three and four**, which is the pattern worth naming:

| Finding | What my earlier fix missed |
|---|---|
| Preserve the landed transcript boundary (P1) | Moving rolling-summary/scene-break scheduling server-side dropped the `upto` those routes take deliberately — without it a fold counts an unanswered prompt as covered and never self-repairs, and scene break is asked about a beat whose reply hasn't arrived. The runner captures the completed turn's length before releasing its slot. |
| Reject retries whose review is already stale (P1) | Requiring a review to *exist* still accepts a retry against one the transcript has outgrown: full budget spent, merged into the old payload, guaranteed refusal afterwards. Pre-flight validates the watermark. |
| Disambiguate concurrent drafts of the same kind (P1) | "Adopt an in-flight run of this kind" is ambiguous once drafts overlap on a coarse subject — two image descriptions in one world, and a remounted editor adopts the wrong image's text. Recovery keys on the attempt id. |
| Recover expired drafts without refetching a scene (P2) | `run_gone` → "refetch the scene" was written when every run was scene-scoped. World, campaign and global drafts have no scene, and an opener isn't in a transcript until accepted. It now resolves by class: turns and reviews refetch, drafts offer regeneration. |

---

**On convergence.** Five rounds, 44 findings, all resolved. The counts (10, 11, 8, 8, 7) look flat, but the composition has changed decisively: rounds 1–2 challenged the core design, rounds 3–4 were scope-expansion fallout, and round 5 is four self-inflicted regressions plus three findings that are really about implementation facts — the kind the plan and implementation gates exist to catch, and which a spec can only guess at.

That is my signal to stop specifying. Remaining review effort is better spent on a plan that has to name the sync/async decision, the migration, and the consumer adaptations concretely.

---

charlesmsiegel:
Round six: three findings, verified, all three hold. Addressed in `930e385`. All three are corrections to **my own** earlier fixes rather than to the original design.

**`Preserve scheduling after non-turn transcript mutations` (P2)** — round four moved the rolling-summary and scene-break triggers to the turn runner so a locked phone couldn't drop them. Right for turns, wrong as a *replacement*: `askAfterPost` has **eight** call sites and only one is the end of a generated turn. Edits (`CampaignView.tsx:2185`), retcons (`:2230`), cuts (`:2317`), manual rolls (`:2543` — *"a roll is a post too"*), checks (`:2595`) and take swaps (`:3351`) all fire it, and none involves an LLM run. Routing everything through the runner would have dropped seven of eight, leaving a stale summary and a scene break never asked about. The runner now covers the turn case — the one the client cannot serve with the phone locked — and the client keeps the rest.

**`Do not adopt a conflicting run as the rejected send` (P1)** — two of my fixes contradicted each other. The round-one 409 contract told a rejected caller to attach to the named run "rather than surface an error," but a rejected send never had its prompt appended, so the loser of two concurrent tabs would render *another tab's reply* while its own prompt went nowhere — precisely the "did my send land?" ambiguity the round-three attempt id was introduced to end. The 409 now exposes the run for busy-state display and for a client that recognises its **own** attempt id; anything else keeps its prompt and reports the scene busy.

**`Route completion taps by immutable scene identity` (P2)** — a notification outlives the moment it was posted. The scene may have been deleted (which this design supports, and which is exactly when the *error* notification fires) or renamed, changing the `sid` while the notification sits in the drawer. Taps now carry the immutable identity and resolve it at tap time, with a campaign-level fallback.

---

**Six rounds, 47 findings, all resolved.** Counts: 10, 11, 8, 8, 7, 3. Round six is the first that found nothing wrong with the design itself — every finding was a consequence of a fix from an earlier round, and the count fell by more than half.

That is convergence, and it is where the spec stops. What remains is genuinely implementation-shaped: whether the six synchronous handlers become `async def` or bridge onto the loop, how the scene-identity backfill sequences against the existing startup migrations, and the order of the frontend consumer migration. Those belong in the plan, which lands on this same branch.

---

charlesmsiegel:
Round seven: five findings, verified, all five hold. Addressed in `f407a46`. **Four of the five were consequences of fixes from earlier rounds.**

**`Account for the scene-identity snapshot change` (P2) was the most valuable**, because it caught a false instruction rather than a design flaw. The spec said `snapshot.json` should not move and that a moved snapshot "is a finding, not a regeneration." That was written before round three added the scene identity field and left standing after. `sweep.py:188` snapshots the whole `store.scenes.read_scene` payload, so the snapshot *will* move for every scene in the frozen fixture — and the instruction as written would have sent an implementer chasing an expected gate failure as a bug, or quietly leaving the backfill out of the compatibility sweep. Now: the sweep runs the backfill, the diff is reviewed to confirm the new field is the only change, and it is regenerated deliberately per `AGENTS.md`.

**Two races the earlier fixes half-closed:**

- **`Reserve against every transcript mutator` (P1)** — reserving before the snapshot closed the gap *before* an absorb; edits, retcons and cuts still rewrite the transcript *during* one. `saveEdit` (`CampaignView.tsx:2163`) and `saveRetcon` (`:2204`) gate on `rolling`, not `sceneLocked`. Now one rule: while a `turn` or `review` holds a scene, that scene's shape does not change — renames, repads and transcript mutators refused together.
- **`Invalidate terminal run results when cancelling a review` (P1)** — Cancel deleted the pending file but left the landed run pollable for ten minutes with its result attached, so mount discovery could render the dismissed review; and since `PUT /chronicle` consumes the payload and token rather than the file, it could still be **saved**. Cancel now tombstones the run.

Plus: mount discovery was reusing the lost-response lookup (a fresh mount has no cursor and its scene fetch already holds the reply, so replaying from `0` duplicates it and `next_index` yields nothing) — discovery now attaches only to live runs; and `cancelled` runs no longer notify, since only success/error text exists and a deliberate Stop would have posted an error.

---

## Stopping the spec-review loop here

Seven rounds, 52 findings, all verified and all resolved. Counts: 10, 11, 8, 8, 7, 3, 5.

The count is no longer the useful signal — the composition is. Rounds one and two challenged the design. Rounds three and four were fallout from widening the scope. Rounds five through seven have been **overwhelmingly findings about my own previous fixes**: 4 of 5 this round, 3 of 3 last round, 4 of 7 before that. Each push changes text, and the next review reviews the changed text. That is a loop that pays diminishing returns and does not terminate on its own.

Every remaining risk I can name is implementation-shaped and cannot be settled by more prose: whether the six synchronous handlers become `async def` or bridge onto the loop, how the identity backfill sequences against `migrate_scene_ids` and `bake_char_macros`, and the order of the frontend consumer migration. Those get decided in the plan and *tested* in the implementation, where there is code to run rather than text to re-read.

So this spec is done. Further review effort moves to the plan, which lands on this same branch.

---

PR #407, opened 2026-08-22
Spec: turn scope and PC authorship

charlesmsiegel:
Spec only — no implementation. `docs/superpowers/specs/2026-08-21-turn-scope-and-pc-authorship-design.md`.

## Problem

Two complaints, one root cause. Every reply tries to be a complete scene — establishing narration, a beat run to resolution, a closing image — so nothing breathes. And the model writes the player's character, a gap currently patched in a personal global system prompt.

They're the same bug. The scene prompt specifies **shape** in exhaustive detail (five length knobs plus an adaptive drift corrective) and specifies **scope** nowhere. Nothing tells the model what one reply is relative to the scene it sits in. Three things then conspire:

- `response_format.j2` has no placement rule for `**Grimoire:**`, so the model bookends — and that bookend *is* the shape of a scene.
- `response_budget.j2` hands over a word target with no scope statement, which reads as "fill this with a satisfying unit of story".
- `opener_shape.j2` prescribes exactly that bookended shape, and the opener is the first assistant message in every transcript — a few-shot exemplar nothing ever contradicts.

The closing narration block is also where PC authorship lands, since a block that has to "land" the post gets pulled toward summarizing the moment's effect on the POV character. One rule addresses both.

`scene_break/system.j2` already articulates the missing concept ("movement in the middle of an unresolved beat is pacing, not a boundary") — but only ever shows it to the judge that decides whether a scene ended, never to the writer.

## What the spec proposes

- **`turn_scope`** — a new lock-in section (standard / pcless variants), `except_opener`.
- **`player_character`** — a new lock-in section enumerating the real failure modes. The line is volition: sensation is the world's, what they make of it is theirs; reflex and recognition, never conclusion.
- **`response_format`** — drops the old trailing PC clause, adds "the last speaker block is a character's".
- **`/end`** — a typed command that vanishes from the transcript and licenses the next reply to close the scene.
- **Precedence** stated in the blocks themselves, `natural_prose.j2`'s adversarial-review finding applied to three always-on blocks instead of one.

Shipped in two phases. Phase 1 is prompt-only and answers both complaints without `/end` existing; Phase 2 is the command. The order is forced — the wrap variant overrides a section that must exist first.

## Review history

Eight adversarial passes; the design changed materially under four of them.

- **`/end` was unreachable in both of its commonest invocations.** The flag was keyed to the index of the player post carrying it, but `ephemeral = is_pcless or not content.strip()` means neither a bare `/end` nor any pcless scene appends a post. Replaced with a scene-level boolean.
- **The PC boundary claimed absoluteness from the weakest slot in the message list.** A card's `post_history_instructions` lands after the whole system prompt *and* the whole transcript — the slot built to outrank. It now also rides there (static, no detector) and is non-removable in the layout editor, which `pack.LOCK_IN` never made it.
- **A three-way collision on "how a reply ends"** between the roll fence, the transient tracker, and the new rule. Scoping to *speaker blocks* lets all three be true; `transient_tracker.j2`'s "after the last line of narration" needed rewording since it presumes the ending now forbidden.
- **`turn_scope` rendered into openers**, contradicting `opener_shape.j2` — the message sent last specifically to outrank the system prompt.

Also caught: `evals/cases.py:203` renders `response_format.j2` with the var being removed and without the two being added, so it raises under `StrictUndefined`; cassettes match on `system_contains` and raise rather than defaulting; and "rewind clears" / "regenerate persists" coexist only because regenerate uses `remove_trailing_assistant_run` rather than `delete_from` — now written down as a load-bearing invariant with its guard test named.

## Notes

- No code changes, so no `make check` run — the diff is one markdown file.
- Codex gates were explicitly skipped for this spec at the author's request; `/brutal-review` stood in.
- The section prose in §1 and §2 is the actual deliverable and is the part most worth reading closely — it's cheaper to edit now than after it's wired into evals, cassettes and the frozen snapshot.

---
_Generated by [Claude Code](https://claude.ai/code/session_0111MGyKdV21sdRQ4LFM6vBm)_


charlesmsiegel:
All five Codex findings verified against the code and addressed in c1c9965. Every one held up; two invalidated stated reasoning rather than just wording.

**P1 — closing rule unsatisfiable with no NPCs.** Confirmed. `backend/tests/test_context.py:466` is a non-opener scene with a player and zero NPCs, and its own comment says so. There the only character present is the player's, whom §2 forbids writing, so requiring a character block last leaves narration the only possible output *and* forbids it. The likeliest escapes — invent an NPC, or write the PC — include the exact failure this spec exists to prevent. Now gated on an NPC being in the scene, with the grader taking the same context.

**Snapshot before allowing live cancellation.** Confirmed, and this one broke my justification, not just the wording. I had argued the route was safe to leave unguarded because a live turn has already assembled its prompt. It hasn't: `_chat_run` closes its setup lock at `:525` and composes at `:528`/`:549`, so a cancel in that window clears the flag before `_assemble` reads it. Fixed by snapshotting under the hold that already exists and passing it through `_turn_override(turn)` — the per-turn override `_assemble` already accepts — so composition reads the snapshot rather than the file. No new guard needed.

**Ephemeral failures leave the flag set.** Confirmed. `undo_user_post=` is wired only into the persisted-post branch at `:571-574`; the ephemeral branch at `:526-548` builds its stream with no undo at all. Since a bare `/end` scrubs to `""` it is *always* ephemeral, so the commonest invocation was also the one with no cleanup. The failure clear is now its own hook invoked from both branches.

**Non-removability must reach the editor.** Confirmed. `layout.describe` returns `{id, label, default_label, tier, enabled}` and `PromptLayoutEditor.tsx:61-63` renders the checkbox with no `disabled`, so a user could untick the section, save, see it reported off, and get it anyway — the UI stating the opposite of what generation does. `removable` now propagates through `describe`, the API type and the editor, with a frontend test.

**post-history line dropped half the boundary.** Confirmed, both directions. The short form omitted conclusions, wants and presumed answers — leaving "describe what the player realizes" unrebutted in the one slot positioned to rebut it — while banning "action" unqualified, which contradicts §2's permission for involuntary reflexes. Rewritten to carry the full volition-versus-sensation distinction.

Still spec-only; no implementation in this PR.

---

charlesmsiegel:
Second round: five more findings, all five verified and fixed in the latest push. Two were regressions introduced by the previous round's own fixes.

**Route-level cleanup, not a stream hook.** Confirmed. I'd specified the failure clear as "its own hook invoked from both branches" — but a hook that only exists once a stream does misses every way setup can fail after `wrap_next` is set: `compose_turn`/`compose_director_turn` raising, a template failing, and `_chat_stream`'s synchronous claim, which the route's own comment describes as raising `StoreBusy` "at this line" with the route answering 409 having sent nothing. Moved to route level around the whole post-mutation span, and the tests now cover a composition failure and a claim failure rather than only a streaming error.

**`post_roll_proposal` needs the snapshot too.** Confirmed, and this was a genuine gap in my scope rather than a wording problem. `routes/mechanics.py:_continuation_messages` and `_declined_continuation_messages` call `compose_turn` with **no `turn=` override at all** (`:98`, `:105`), so a wrap reply that stops at a roll fence and is then accepted or declined composes its continuation straight off the file — the identical cancel race, one route over. CLAUDE.md inventories `post_roll_proposal` as one of the five detached-run handlers, which is exactly where I should have looked. Both builders now take the snapshot, with a race test on that path.

**`turn_scope` never said a style can't relax it.** Confirmed, and the sharpest of the five. The Precedence section claims these relationships are "stated in the blocks themselves", and the block didn't state this one. The follow-on point is the real reason it matters: section order can't serve as the mechanism either, since `layout.py` lets a user reorder freely, so `turn_scope` sitting after `prose_style` is a default, not a guarantee. Added as a bullet to both the `standard` and `pcless` variants.

**Plural in the post-history guard.** Confirmed — I fixed plural rendering for `player_character.j2` two rounds ago and then wrote this one singular-only. In the one slot positioned to override a hostile card instruction, a second seated PC left unnamed is unprotected precisely where it matters. Mirrors the section's singular/plural handling now, with a multi-PC test.

**Phase 1 "no frontend".** Confirmed, and self-inflicted: last round's `removable` fix put the layout editor into Phase 1 while the phase summary still said no frontend work. A plan following the summary would have skipped the UI half and shipped the misleading checkbox the fix exists to remove. Corrected to say Phase 1 carries no new routes and no new stored state, but does carry that editor work.

---

charlesmsiegel:
Third round: five more, all verified, all fixed. The P1 and one P2 are both defects in the *previous* round's fixes rather than in the original design — worth saying plainly.

**P1 — roll requests. My last fix was wrong, not merely incomplete.** I claimed that scoping the rule to *speaker blocks* let the roll fence, the transient tracker and the closing rule all be true at once. That's right for the tracker, which `_persist_reply` strips before the reply is split into posts. It is wrong for rolls: `mechanics_response_format.j2` asks for the fence *"mid-narration, at the moment of the attempt, then STOP writing immediately after the closing fence"*, so in a compliant roll request the enclosing speaker block **is** `**Grimoire:**` and the reply ends there by design. And when the PC is the one attempting the check, §2 leaves no permissible character block at all. Both the rule and the grader now carry an outright roll-fence exemption, with detection reusing `store/fence.py`'s `OPENER` rather than a second copy of the grammar.

**The carrier I picked cannot carry it.** `_turn_override` returns a dict built from the typed `ResponseSettings` model (`routes/common.py:254`), and `_assemble` hands it to `response_presets.resolve`, which walks only `("style_id",) + lengths.KNOBS` (`response_presets.py:251`). A `wrap` key there is silently discarded and composition keeps reading live frontmatter — so my "fix" for the cancellation race would have left the race intact and harder to spot. Replaced with an explicit `wrap: bool | None` keyword on `compose_turn` / `compose_director_turn` / `_assemble`, separate from the response-preset path.

**`post_retry` was missing from the snapshot inventory.** Same lock-then-compose shape — lock closes at `:619`, composes at `:621-622`. I added `post_roll_proposal` last round and still didn't sweep the rest of the handler list.

**The indicator's copy is false after a successful wrap.** Direct consequence of my own lifecycle: the flag deliberately stays set so re-rolls reproduce the closing reply, so "the next reply will close this scene" stops being true the moment the wrap lands — the next ordinary send clears it and resumes play. `wrap_next` is now two-valued (`pending` / `consumed`) with copy per state. Stored rather than derived, because a bare `/end` appends no post, so the transcript's last message can't distinguish the two.

**Disabling `turn_scope` wasn't a real opt-out.** The structural half lives in `response_format`, a separately registered section, so a user would have had to disable the reply-format contract — losing the `**<Name>:**` markers every downstream split depends on — to opt out of pacing. The closing-narration rule now renders only when `turn_scope` is enabled. The PC boundary doesn't ride on that gate; it stays non-removable.

---

Issue #54, opened 2026-07-11
Library sharing: export / import world bundles (zip a world directory)


charlesmsiegel:
Part of the **Worlds & Library Content** milestone. From the pre-rebuild target feature list:

> Library sharing: export / import world bundles (zip a world directory)

## Current state
No world export exists. The precedents are close by: `backend/src/grimoire/store/epub.py` builds a zip in memory with stdlib `zipfile` and `GET /campaigns/{cid}/export.epub` (`routes.py`) returns it as a `Response` with a `Content-Disposition: attachment` header (deliberately declared before the generic `/{kind}` routes); `store/cards.py` already parses zip uploads (CHARX) arriving via multipart `UploadFile` (`POST /worlds/{wid}/characters/import`).

A world directory is fully self-contained: `world.md`, `locations/`, `lore/` (+ per-entity `assets/`), `greetings/` (+ per-greeting `assets/` and `subjects.json`), `characters/<cid>/` (card JSONs, `assets/<vid>/`, `tagline.md`), `pcs/`, `plotmap.json`, `tags.json`, `calendar.json`. One real snag: localized images are referenced by **absolute URLs embedding the world id** — `store/localize.py` writes `/api/worlds/{wid}/characters/{cid}/versions/{vid}/images/{name}` and `/api/worlds/{wid}/greetings/{gid}/images/{name}` into card/greeting text, so an import under a new wid must rewrite them.

### Option A — Whole-directory zip round-trip (recommended)
- `GET /worlds/{wid}/export.zip`: walk `worlds.world_root(wid)` and zip everything, paths relative to the world root (declare the route before `GET /worlds/{wid}/{kind}` so `export.zip` isn't captured as a kind).
- `POST /worlds/import` (multipart): read `world.md` from the archive for the name, mint `wid = uniquify(slugify(name))` (`store/paths.py`), extract, then rewrite `/api/worlds/<old-wid>/` → `/api/worlds/<new-wid>/` across `.md` and `.json` text files.
- Frontend: Export button per world card in `WorldsView.tsx` + an import file input.
- Recommended: the store layout *is* the format; least code, lossless.

### Option B — Versioned manifest bundle
- Add a `bundle.json` (format version, world name, app version) at the zip root; import validates known kinds, skips unknown files, and can migrate older bundles.
- More forward-compatible for sharing between app versions; more code now.

### Option C — Document manual folder copy
[...]

Issue #351, opened 2026-08-15
A second CampaignView.test.tsx race reds CI on PRs that never touch the file


charlesmsiegel:
`2b13ae9` ("The frontend flake was never load: a never-settling mock outliving its test") found and fixed a real defect in `CampaignView.test.tsx`. Something else in that file still fails intermittently, and it has a **different signature**. That commit predicted this outright:

> Not claimed to be the last of it. Under artificial CPU saturation the suite still loses tests to other races — one of them an assertion failing at 280ms, which is a different defect.

This issue records the measurements so the next person does not start from theory. It was found while driving #347 to green; that PR does not touch this file.

## What is different from the leak that was fixed

| | leak fixed in `2b13ae9` | what remains |
|---|---|---|
| failure mode | `findBy*` timing out at the async ceiling | **sync** query failing immediately (`getAllByTitle`, `getByLabelText`) |
| meaning | a promise nobody resolves; the element never arrives | the element genuinely was not rendered at that moment |
| responds to a higher ceiling | no (still failed at 8074ms with an 8000ms ceiling) | untested |

A sync query failing means the preceding `await` resolved and the component rendered *without* the control the test expects. That is not a hang.

## Reproduction

Measured on the tree CI actually builds — a PR head merged with `main` — on a 4-CPU container:

| run | result |
|---|---|
| `main` alone, full suite ×2 | green (1488/1488) |
| `CampaignView.test.tsx` alone, merge tree | green (322/322) |
| merge tree, files **parallel** ×5 | fails ~1 in 3, **a different test each run** |
| merge tree, `--no-file-parallelism` | green (1506/1506) |

Sequential green + parallel flaky + a different victim each run is contention. The merge tree differs from `main` only by ~18 extra tests, i.e. by load.

## Observed victims

From CI (run [31859430621](https://github.com/charlesmsiegel/grimoire/actions/runs/31859430621), both in one run):

- `declining the cut's confirm does nothing` — `Unable to find a label with the text of: Delete message 2 and everything after it`
- `a manual dice roll's transcript line has no Edit control` — `Unable to find an element with the title: Edit message` (`CampaignView.test.tsx:651`)

From local parallel runs, one per failing run:

- `a reroll cannot regenerate a scene whose transcript is not the one on screen`
- `cancelling keeps a one-shot response override for the retry`
- `a failed voice check is reported, never silently swallowed (#59)`

## Leads, not conclusions

[...]

PR #414, opened 2026-08-22
Fork a world (deep copy) — and the import that could delete a directory outside the store

### LINKED ISSUE #41: Fork a world (deep copy)
Part of the **Worlds & Library Content** milestone. From the pre-rebuild target feature list:

> World fork (deep copy)

## Current state
No world fork exists — `store/worlds.py` has only `create/read/rename/delete_world`. The established deep-copy precedent is world→campaign copy-on-create: `store/campaigns.py:create_campaign` copies every synced entity file **and its assets tree** (`entities.synced_refs` + `shutil.copytree` of `<kind>/<eid>/assets`), `plotmap.json`, whole character/PC directories, and the calendar (`calendars.copy_calendar`), while recording base hashes into `sync.md`.

A world directory (`worlds/<wid>/`) is fully self-contained: `world.md`, `locations/`, `lore/`, `greetings/` (+ per-greeting assets and `subjects.json`), `characters/<cid>/` (cards, versions, assets, `tagline.md`), `pcs/`, `plotmap.json`, `tags.json`, `calendar.json`. Nothing inside a world references its own `wid`; only campaigns point at worlds (`campaign.md` frontmatter `world`), and a fresh fork has no campaigns — so a directory copy is referentially clean.

### Option A — whole-directory copy (recommended)
- `store/worlds.py:fork_world(src_wid, name) -> str`: `wid = uniquify(slugify(name), ...)`, `shutil.copytree(world_root(src_wid), world_root(wid))`, then rewrite `world.md` (`name`, fresh `created`/`updated` via `now_iso()`).
- Route: `POST /worlds/{wid}/fork` with `NameBody` — **must be declared before** the generic `POST /worlds/{wid}/{kind}` in `routes.py`, or "fork" is captured as an entity kind.
- Frontend: a "Fork" action on the `WorldsView.tsx` card row (next to rename/delete), prompting for the new name; `api.forkWorld` in `api/client.ts`.
- Recommended: worlds are self-contained directories, so copytree is complete by construction — new files added to the world layout later are forked for free.

### Option B — ref-walk selective copy
- Reuse the `create_campaign` walk (synced refs, plotmap, actor dirs, calendar, tags) with per-kind include flags ("fork without greetings", …).
- Enables partial forks, but every future world file must be added to the walk by hand (the exact trap `ensure_campaign_copy` exists to patch).

### Option C — fork via export/import bundle
- Implement #54's zip export, then fork = export + import-under-new-name. One copy engine for both features, but blocks fork on a bigger feature.

## Implementation notes
- Do the copytree into a temp sibling and rename into place, or clean up on failure, so a crashed fork doesn't leave a half-world that `list_worlds` would happily list (it only requires `world.md` to exist).
- No id rewriting is needed: entity/greeting/character ids are world-relative and stay valid; internal image URLs in localized bodies are `/api/worlds/{wid}/...` — **check for absolute self-references**: `store/localize.py` rewrites card/greeting image refs to URLs containing the source `wid`, so forked bodies may point at the old world's routes. Either rewrite those strings during fork (simple text substitution of `/worlds/<src>/` → `/worlds/<new>/`) or accept stale-but-working links while the source world lives; the rewrite is cheap and correct — do it.
- Tests: `backend/tests/test_worlds.py` — fork a world containing an entity with assets, a character with versions + tagline, plotmap, tags, calendar; assert independence (edit fork, source unchanged) and the URL rewrite. `monkeypatch.setenv("GRIMOIRE_HOME", tmp_path)`.

## Dependencies & related
- #54 — world bundle export/import shares the "world dir is the unit" insight (Option C unifies them).
- #72 — campaign fork is the campaign-side sibling; keep naming symmetrical.
- #38 — fork should carry genre/description automatically (it does under Option A, since they live in `world.md`).

charlesmsiegel:
Closes #41.

## What this adds

`POST /worlds/{wid}/fork` deep-copies a world into a new one, driven by a `⑂` action on each world card.

A world directory **is** the world, and nothing inside it names its own id except the serving URLs `store/localize.py` writes into card and greeting text. Campaigns are the only records that point *at* a world, and a brand-new fork has none — so a whole-directory copy plus a URL repoint is referentially complete **by construction**, and a part added to the layout next month is forked without anybody editing this code. That is the trap `create_campaign`'s hand-written walk keeps falling into, and the reason this is `copytree` and not a per-kind walk.

Only three things are rewritten afterwards, all of them identity: the name and both timestamps. `created` because the fork started existing now; `updated` because `list_worlds` sorts on it, and a fork of a world last touched two years ago would otherwise land at the bottom of the shelf the user is looking at for it.

## The shared seam

The two hard parts were already solved next door by the bundle import (#54), so they moved out of `world_bundle.py` into `store/worlds/staging.py`, which import and fork now share:

- **Nothing half-made is visible.** `list_worlds` calls any directory holding a `world.md` a world, and both features spend real time with that file already written and the copy still running. The tree is built where the listing never looks and enters the library by a single `rename`.
- **A world's records name the world.** `publish` owns the id-race retry and re-points on each attempt, so the world it publishes always references itself.

`atomic.is_write_temp` moved the same way — it recognizes the temps `atomic` generates, so the pattern belongs beside the code that writes them rather than in one of the two walks that skip them.

## A pre-existing bug this turned up

Reading `import_bundle` to check that the contract the new module claims is one *both* callers keep — it was not.

`import_bundle` held its work directory and the world's slug in **one name**: `base` was the staging `Path` until `_world_name` was read, and the slug from then on. Its `finally` then ran `shutil.rmtree(base)` on a bare relative name, resolved against the **process working directory**.

Reproduced before fixing: with a shell sitting in a directory holding a `saltmarch/` of the user's own, importing a world called "Saltmarch" **deleted it** — files grimoire never wrote, outside the store entirely — and leaked the staging tree on every import, successful or not.

mypy had this the whole time (`expression has type "str", variable has type "Path"`, plus two `arg-type` follow-ons), sitting in `lint-baselines/mypy.json` as accepted debt. That baseline is three findings smaller now.

The fix is not the rename, though: both writers kept their work directory by hand and one got it wrong, so that shape is gone behind a `staging_tree()` context manager that owns both halves.

## Other findings from the review rounds

- **A read-only record sank the whole fork.** `copytree`'s default `copy2` carries mode bits, so a record the user had chmod'ed `0444` — a protection `store.atomic` deliberately honours — arrived read-only in staging and the URL repoint could not rewrite it. `copy_function=shutil.copyfile` copies content and no metadata; the fork's files are new files, and the store keeps every timestamp it cares about in its own frontmatter.
- **A failing refresh after a successful fork said nothing at all**, which is indistinguishable from a fork that never happened — and the retry it invites makes a second copy of a gigabyte.
- **The entity-with-assets case #41 asked for by name was not covered.** Entity kinds get the same per-version asset folder characters do (`store/assets.py`), and the seed gave galleries only to a character and a greeting — so the whole-tree test would have caught a copy that dropped them, but only if the seed had had one to drop.
- `.row-actions` had no disabled styling, because this is the first disabled button to appear in one.

## Tests

`backend/tests/world_fixtures.py` holds the seeded world both this suite and the bundle suite copy, so the two cannot drift into testing different amounts. It now covers all three localized-URL shapes (character version, greeting, PC persona), an entity's own gallery, the `descriptions.json` sidecar that carries prose, and the `subjects.json` sidecar that carries ids and no URL.

- `test_world_fork.py` — whole-tree equality minus the files a fork is *supposed* to change; assets byte-verbatim; URL repointing across every character version, a fork of a fork, and an id that prefixes another; independence in both directions and after the source is deleted; a symlink becoming a real file; a failure publishing nothing; a lost id race refusing rather than merging; write temps skipped and a look-alike `.notes.tmp` kept.
- `test_world_staging.py` — the seam pinned directly, since it exists to make a mistake unrepeatable.
- `test_world_bundle.py` — two regression tests for the cleanup bug, both verified failing against the old code.
- `test_atomic.py` — `is_write_temp` asserted against names the module *really produced*, captured mid-write, rather than a hand-written example that would drift the same way.
- `test_routes.py` — the route end-to-end, including that a forked localized image URL actually **serves** over HTTP, and that declaring a literal segment before the `/{kind}` catch-all did not shadow it.

Mutation-checked: removing the repoint, the identity stamping, or the id canonicalization each reds the test written for it, and a copy that skips every `assets/` directory — the shape a hand-written per-kind walk degrades into — reds five.

## Deliberately not done

- `module_edit/packs.py:duplicate_module` has the same read-only-record bug (`copytree` with `copy2`, then `atomic.write_text` on a copied file). One line to fix, but it is a different feature with its own locking and migration machinery and should not ride along here.
- Abandoned staging trees are still never swept — a `SIGKILL` mid-copy leaks a world-sized directory. Recorded as a known limit with the reason a wrong sweep is worse than the leak.

## Verification

`main` was merged in (it had moved 31 commits ahead). On the merged tree: `make check-lint`, `check-mypy`, `check-eslint`, `check-templates` all at baseline; backend **6877 passed**, frontend **2030 passed**.

One pre-existing failure is deselected when running locally as root: `test_a_read_only_record_is_not_silently_replaced` needs `os.access` to mean something, and for root it does not. Confirmed failing on unmodified `main` too, and green on both CI backend legs, which do not run as root.

https://claude.ai/code/session_01BRbPS558tQa7nU7WnkZSsJ


charlesmsiegel:
## Codex findings: what landed, and the three I did not take

Eight of eleven findings are fixed in `8902164` and `44fdadd`, with a regression test each. Three are left, deliberately, and here is why plus what I would do about them.

### Not taken — needs a decision, not a patch

**P1 · Make long-running forks idempotent across lost responses.** Correct: a dropped response after publication makes the client report a definite failure, and the natural retry copies a gigabyte again. The fix is an idempotency key, and this repo already has the shape for it — `store.attempts` is exactly "was this send's post already applied", recorded beside the write and cleared on rollback. Proposed patch: have the client mint an operation id, send it as a header the way `X-Grimoire-Attempt` is sent for turns, and have `fork_world` record it beside the published world so a retry with the same id returns the existing fork instead of making a second one. That is a new durable record and a new header contract, which is more than this PR should decide on its own.

**P2 · Keep the fork pinned to one storage root.** Correct. `staging_tree()` captures its work directory, so cleanup is safe, but `publish()` re-resolves `home()` for its destination — so a storage-location change mid-copy publishes into the new library, or fails and is reported as an id conflict. Proposed patch: resolve the worlds directory once per operation and thread it through `staging_root`/`publish` instead of each of them calling `home()`. That changes the seam's signature for all three callers (`create_world`, `fork_world`, `import_bundle`); the alternative — having a world copy participate in the `runs_in_flight` exclusion `PUT /config/data-dir` already uses — needs a registry worlds do not have. Both are worth doing; neither is a small edit, and the import has carried this since #54.

**P2 · Same-second sort ties.** Partially taken. `list_worlds` now breaks a tie on `created`, which fixes the case you actually hit (an old world edited just before you fork it). Two worlds *created* in the same second still tie on both keys, and closing that needs sub-second `now_iso()`, which every stamp in the store and the frozen campaign's snapshot are written against. The residual is stated in `worlds/read.py` rather than left implied.

### Unrelated to this PR, but reproducible

`CampaignView.test.tsx > a player post says what answering it cost, over every reroll` (added in the cost work on `main`, not by this branch) fails under full-suite load — 2 of 2 local runs here, and it red-flagged this PR's first CI run before base was merged. It passes in isolation and on a quiet runner, which is the #351 shape.

Diagnosis: the cost effect in `CampaignView.tsx` calls `setPostCosts(null)` on every `[cid, activeId, ctxKey]` change, so the chip blinks out and back on each pass; under load the `findByText` window can close first. The clear looks redundant — `postChips` already gates on `postCosts.cid === loaded.cid && postCosts.sid === loaded.sid`, so a stale chip cannot render anyway.

Removing those two lines makes the full suite pass (2032/2032, against 2 failures out of 2 without). Not pushed here: it is a behaviour change in a feature this PR does not touch, and belongs to whoever owns the cost work.

---

charlesmsiegel:
## Correction: I did push the `CampaignView` cost-chip change after all

In the comment above I said I would leave the flaky cost test to whoever owns that work. That was the right call on scope and the wrong call on consequences — the same test then red-flagged this PR's CI a **second** time (two of four runs now, always `CampaignView.test.tsx:122`), and a PR that flips red at random cannot be landed on. `06432ea` pushes the fix.

To be clear about what it is and is not: the test still runs and still asserts exactly what it did. Nothing is skipped, disabled, or quarantined.

**What changed.** The post-cost effect cleared `postCosts` to `null` on every `[cid, activeId, ctxKey]` change before refetching. That clear bought nothing — `postChips` already refuses to render a bucket whose `cid`/`sid` is not the transcript on screen, so a stale chip could never have been shown. What it did buy was a window with no chip in it, and under load the test's `findByText` can open and close inside that window.

**Evidence, not inference:** full suite fails 2 of 2 local runs with the clear, passes 2032/2032 without it; CI failed the same single test at the same line on two of four runs.

**The one real behaviour change, and it is yours to accept or reject:** during a context change the previous cost figures now stay on screen until the new ones arrive, instead of vanishing first. I think that is better — it is the same data for the same scene — but if you would rather keep the blink, the alternative is to make the test wait for the refetch explicitly, and I am happy to do that instead.

---

Issue #240, opened 2026-07-28
_safe_id path guard is missing from campaign/world/character root resolvers


charlesmsiegel:
Surfaced by the codebase atlas (`docs/codemap.html`, generated at 81b61ff1).

## Problem

The codebase has a defense-in-depth guard against ids escaping their directory:

```python
# backend/src/grimoire/store/entities.py:37-39  (also scenes.py:98-100)
def _safe_id(eid: str) -> bool:
    """Reject ids that could escape the kind directory (defense in depth)."""
    return eid not in ("", ".", "..") and "/" not in eid and "\\" not in eid
```

It is used in `entities.py` (5 sites), `scenes.py` (24), `modules.py` (2) and `module_edit.py` (3) — and is **absent from the three highest-traffic root resolvers**, which join a caller-supplied id onto a path unchecked:

```python
# backend/src/grimoire/store/campaigns.py:23-24
def campaign_root(cid: str) -> Path:
    return _campaigns_dir() / cid

# backend/src/grimoire/store/worlds.py:27-28
def world_root(wid: str) -> Path:
    return _worlds_dir() / wid

# backend/src/grimoire/store/characters.py:37-38
def _char_dir(root: Path, cid: str) -> Path:
    return _chars_dir(root) / cid
```

## Honest assessment of exposure

**This is not a demonstrated exploit.** FastAPI path parameters do not match `/`, so the obvious traversal vector is blocked by the router before it reaches the store. The practical risk today is low.

The point is structural: the guard was written *because* relying on the caller is not good enough, and these three functions currently rely on the router instead of on the store. Any future caller that isn't an HTTP path parameter — a request body field, a CLI script, a batch importer — gets no protection. `backend/scripts/ingest_scene.py` already takes ids from outside the router.

## Suggested fix

Apply the same `_safe_id` check in `campaign_root`, `world_root` and `_char_dir` (and consider lifting the duplicated helper into `store/paths.py`, since it is currently defined four times).

Issue #144, opened 2026-07-11
Retry-with-backoff + fallback route on exhaustion


charlesmsiegel:
Part of the **LLM / Providers** milestone. From the pre-rebuild target feature list:

> Retry-with-backoff + fallback route on exhaustion

## Current state

There is no retry logic anywhere in the codebase. `backend/src/grimoire/openrouter.py:OpenRouterClient` makes exactly one HTTP attempt per call: `.stream()` opens `http.stream("POST", API_URL, ...)` once and either yields deltas or raises `OpenRouterError(kind)` on failure (`kind ∈ missing_key|auth|rate_limit|network|bad_response`, set by `_status_kind()` — 401/403→`auth`, 429→`rate_limit`, else `bad_response`; any `httpx.HTTPError` or other exception →`network`). `.complete()` just joins `.stream()`. The `_client()` httpx client uses `httpx.Timeout(120.0, connect=30.0)` — configured, but nothing catches a timeout and retries; it surfaces as `network`.

Every call site in `backend/src/grimoire/routes.py` (`_chat_stream`/`_ephemeral_stream` for chat/retry/regenerate/opener, and the blocking `client.complete(...)` calls in `post_absorb` — extraction plus one per-NPC dossier loop, `post_scene_suggestions`, `post_character_tagline_generate`) propagates whatever `OpenRouterError` the client raises straight to the response (`{error:{detail,kind}}` over SSE, or a 502 for the blocking endpoints) with no retry wrapper.

There is exactly one provider (OpenRouter, one base URL, one key) and one `model` in `store/config.py:read_config()` — so "fallback route" cannot mean a second provider today; it can only mean a second **model** on the same OpenRouter account.

### Option A — Backoff wrapper around `.complete()` only (recommended)
- Add a `retry()` helper in `openrouter.py` (or a thin wrapper in `routes.py`) that retries `rate_limit`/`network` kinds with exponential backoff + jitter (bounded attempts, e.g. 3), and does NOT retry `auth`/`missing_key`/`bad_response` (not transient).
- Apply it to the five blocking `complete()` call sites (absorb extraction, per-NPC dossiers, scene-suggestions, tagline). These are naturally retryable — nothing has been streamed to a user yet.
- Fallback model: an optional `fallback_model` key in `config.md` (`_CONFIG_KEYS`), tried once after the primary model's retries are exhausted.
[...]

Issue #141, opened 2026-07-11
Multi-provider gateway (cloud + local, plugin-loaded)


charlesmsiegel:
Part of the **LLM / Providers** milestone. From the pre-rebuild target feature list:

> Multi-provider gateway (cloud + local, plugin-loaded)

## Current state

**⚠️ Pre-rebuild concept — needs re-scoping.** This assumes a plugin-loaded provider gateway; the rebuilt app has no plugin system (#182) and exactly one hard-coded provider.

- The only LLM path is `backend/src/grimoire/openrouter.py`: `OpenRouterClient` (httpx) with module constant `API_URL = "https://openrouter.ai/api/v1/chat/completions"`, `.stream()` (OpenAI-compatible chat-completions SSE, yields content deltas) and `.complete()` (joins the stream). Errors normalize to `OpenRouterError(kind ∈ missing_key|auth|rate_limit|network|bad_response)`.
- `routes.py` holds a singleton `_openrouter = OpenRouterClient()` behind the `get_openrouter()` dependency; every call site passes `cfg["model"]` and `cfg["openrouter_key"]` from `store/config.py:read_config()` (one key, one model, default `anthropic/claude-opus-4.1`).
- Re-scoping insight: OpenRouter already aggregates the cloud providers (Anthropic/OpenAI/Google/…), so "multi-provider cloud" is mostly model choice today. The real gap is **local** backends — Ollama, LM Studio, llama.cpp server, vLLM — all of which speak the same OpenAI-compatible wire format the client already parses.

### Option A — Configurable base URL (recommended)
- Turn `API_URL` into config: `llm_base_url` in `store/config.py` (default `https://openrouter.ai/api/v1`), client appends `/chat/completions`.
- Allow an empty key when the base URL is overridden (local servers don't authenticate): relax `routes.py:_require_key` and the `missing_key` guard in `stream()` accordingly.
- Frontend: one text field in `ConfigView.tsx` next to the key input.
- Tradeoff: one active endpoint at a time; switching cloud↔local is a config edit. Recommended: smallest change that covers the actual gap.

### Option B — Named provider profiles
- `providers.json` at the store root (paths via `store/paths.py:home()`): `{pid: {name, base_url, key}}` plus an active selection; new `store/providers.py`, routes `GET/PUT /api/providers`; `get_openrouter()` resolves the active profile.
- Enables per-task provider choice (#142), provider cards (#147), per-provider health (#146).
[...]

PR #417, opened 2026-08-22
Let a single regenerate run against a different model (#77)

### LINKED ISSUE #77: Let a single regenerate run against a different model
**Re-scoped 19 Aug 2026** (see `docs/issue-audit-2026-08-19.md`). The steering half shipped; only the model half is left, and it got cheaper.

## Done — steering hint

`POST /campaigns/{cid}/scenes/{sid}/regenerate` takes `RegenerateBody.guidance`; a non-empty hint renders `templates/scene/regenerate_guidance.j2` and appends it after the rebuilt context. The frontend exposes it as the "Reroll guidance" input in `CampaignView`. The reroll is non-destructive — the previous reply is parked as an alternate under one campaign lock spanning the heal, the decision, the archive and the removal.

## Left — model override

`RegenerateBody` is still `{guidance, response}` (`routes/models.py:126`). There is no per-call model override anywhere.

What changed since this was filed: the original body assumed a single global `config.model` and proposed threading `body.model or cfg["model"]` through. There is now `store/llm_connections.py` — named connections across `openrouter`, `claude` and `openai_compatible` kinds, dispatched through `llm.py` (#141) and picked in `ConnectionsView`. So the override should probably select a *connection*, not just a model string: rerolling against a local Ollama or a different provider is the interesting case, and a bare model id cannot express it.

- [ ] Add the override to `RegenerateBody` — decide connection-id vs. model-string vs. both, given #141's shape.
- [ ] Surface it beside the existing reroll-guidance input, reusing `ModelCombobox` (#149 notes it is designed to be reused as-is at additional call sites).
- [ ] Stamp what actually ran onto the alternate. `SceneInspector` already reads a scene's stamped model to look up its context size, so an alternate generated on a different model must not silently inherit the campaign's.

## Related

- #142 / #143 — per-task and tiered routing. This issue is the one-off manual override; those are the standing policy. Same picker, different scope; neither blocks the other.

charlesmsiegel:
Closes #77's remaining half — the steering hint shipped earlier; this is the model override.

## What it does

The reroll popover gains a route row above the guidance input: a connection picker and a `ModelCombobox`. One reroll, and only that reroll, runs on what they name.

```
[ Default ▾ ][ anthropic/claude-opus-4.1        ]
[ Guide the reroll (optional)…      ][ REROLL ▸ ]
```

`Default` follows whichever connection is active when the reroll is sent; naming a connection pins that one. Choosing a model under Default pins the connection it was chosen against, so a model picked while A was active cannot land on a B that became active in between.

## The decision the issue left open: connection-id, model-string, or both

**Both**, because neither expresses the other's case.

- A bare model id cannot reach a *different provider* — the credentials, base URL and prompt post-processing that makes "try that again on the local Ollama" possible live on a connection (#141), not in a string.
- A bare connection id cannot say "the same provider, its bigger model", which is the cheap everyday case and needs no setup at all.

They compose: `connection_id` alone runs the named connection at its own model, `model` alone runs the active connection at that model, both is the named connection driven at the named model. An absent field is the standing configuration, so an untouched popover posts no body at all.

`_override_connection` returns `(connection, routed)` rather than a sentinel. `routed` means "this ran somewhere it would not have gone anyway" — compared on the *effective* model, so naming the active connection at its own model is correctly not an override.

## Refusals

| Case | Answer |
|---|---|
| model id longer than `MAX_MODEL_CHARS` | **400** — bounded where it arrives, since it reaches three sinks and two cannot be clamped downstream |
| id names no connection (including one the filesystem cannot hold) | **400**, not 404 — this route's 404 means "the scene is gone" and the client acts on it |
| connection cannot send | **409 / `missing_key`**, naming *which* connection |
| unusable override | refused, never silently rerouted |

Resolution happens before the reservation, so a bad body cannot 400 over a scene the route has already left one reply short. The configured fallback (#144) is untouched: an override picks the *primary*.

## Stamping what ran

- **The alternate** is stamped on *every* reroll — it is the only record a variant has of the generation that produced it, and the set exists to be read side by side. `next_model` joins `next_guidance` as one pending pair, spent together onto whatever fills the slot, which is why `disown_guidance` became `disown_pending`. `""` means **no record**. The one documented limit: a generation that records an empty pending pair cannot re-stamp a variant it deduplicates onto.
- **The prompt snapshot** is stamped *only* on an override, for the reason `store.prompt_log` argues at length — its `model` is a field the live context panel reports for the same scene. `None` and `""` are distinct there: `""` is a real answer (a connection with no model configured still generates). This is what keeps `SceneInspector`'s context bar from measuring a 32k local reroll against the campaign's 200k model.

An over-long model id is **dropped, not clipped**: clipping an identifier invents a different one.

## Also in here

- `llm_connections` was outside #240's "never join a caller-supplied id onto a path unchecked" — fine while ids arrived only as URL segments, not once one arrives in a request body. `_read` and `delete_connection` refuse an unsafe id, with store-level tests beside the other resolvers'.
- `GET /llm-connections` now reports `effective_model`, which `GET /config` has always done for the active connection. That deleted a client-side copy of `llm.effective_model`'s rule, a hand-copied `CLAUDE_DEFAULT_MODEL`, and a test that pinned them by scraping a `.tsx` with a regex.
- `api.regenerate` takes its body fields as one object. Two optional fields already cost four branches at the call site; a third would have made it eight.

## Review history

Eleven review rounds — six adversarial self-review passes, three deep-dive rounds, and Codex — surfaced **32 confirmed defects, 9 of them blockers**. Three times, a round's findings were bugs introduced by the *previous* round's fixes, which is why every round here is followed by a full suite re-run rather than targeted tests. Notable catches:

- `None` was carrying two meanings in `_override_connection`, which opened a window where another tab repointing the active connection served an explicitly pinned reroll from somewhere else.
- A `connection_id` past NAME_MAX escaped as a 500 where the route documents a 400.
- A dedup test passed vacuously, hiding a stale model stamp.
- Codex found that a fix described as eliminating a stale window was inert — `getConfig()` answers from a module cache whose own docstring says another tab's writes are invisible to it.

## Verified

- Backend 6973 passed; frontend 2102 passed across 110 files; `tsc -b` clean.
- ruff / mypy / eslint / templates at baseline; pydantic-1.10 (the Android dependency set) suite passed.
- Backend coverage 92.87% against 92.85% on the parent commit — the diff raises it.
- Driven end-to-end against an isolated store with a mocked provider: a routed reroll wrote `"model": "vendor/bigger"` into the alts sidecar beside its `"guidance": "colder"`, recorded `regenerate | vendor/bigger` in the prompt log while the plain chat turn kept the campaign's model, and billed the usage row to the model it ran on. No console errors.

One pre-existing failure is unrelated and reproduces on the parent commit: `test_atomic.py::test_a_read_only_record_is_not_silently_replaced`, which cannot pass when the suite runs as root.

Rebased onto `main` (`b67ef65`).


charlesmsiegel:
Both P2 findings from the previous review are fixed in `e5eb60b`:

- **Reroll route surviving a scene switch** — the popover and its route now clear in the existing `switchingScenes` block, beside the response override, the error banner, the alternates and the replay index. Covered by `switching scenes closes the reroll popover and drops its route`.
- **Stale active connection in the picker** — removed the staleness rather than refreshing it: `RerollRoutePicker` reads `/config` itself on mount (which is only when the popover opens), so `CampaignView` no longer holds the value at all and there is no window in which it can be wrong.

The same commit also carries three findings from an independent adversarial pass: an `id` past NAME_MAX escaped as a 500 where the route documents a 400 (`Path.exists()` raises ENAMETOOLONG, which pathlib does not swallow — the stat now sits inside `_read`'s existing try); the `model`-only path still read the active connection twice, which was the window the previous round's tuple return was meant to close; and `GET /llm-connections` now reports `effective_model`, which deleted a copy of `llm.effective_model`'s rule, a hand-copied `CLAUDE_DEFAULT_MODEL`, and a test that pinned them by scraping a `.tsx` file with a regex.

@codex review

---

charlesmsiegel:
Rebased onto `main` (`b67ef65`). The branch had drifted 57 commits behind; only two conflicts arose, both the same import-list collision in `CampaignView.tsx`, and all seven commits replayed intact. New head is `c3b8bda` and the PR is `mergeable_state: clean`.

Verified on the rebased tree before pushing: `tsc -b` clean, ruff/mypy/eslint all at baseline — with main's post-merge counts (1182/241/935), so the baseline files merged rather than this branch's values overwriting them — and 2099 frontend tests across 110 files passing.

The previous `@codex review` request was against `e5eb60b`, which the rebase replaced, so re-requesting against the head that will actually land.

@codex review

---

charlesmsiegel:
Both findings fixed in `5a3cdd6`, and the first one was a fair hit — the previous round's fix did nothing.

- **`getConfig()` answers from a module cache.** Moving the read into the picker was supposed to remove the stale window, but `api.getConfig` returns `configCache` unless asked otherwise, and that cache's own docstring says it is *"keyed to nothing but this tab's own writes, so a store populated by another tab, another device, or a sync client is invisible to it"*. Relocating a cached read fixes nothing. It now passes `{ fresh: true }`.
- **Pinning the provider.** The active connection is offered in the list again. It and "Default" now mean different things where they previously looked like duplicates — Default follows whatever is active when the reroll is sent, the named row pins that connection — with titles saying so, since the difference only shows when they diverge. And choosing a model under Default pins the connection it was chosen against, so the common path doesn't require the reader to know any of that.

One thing worth stating rather than leaving implicit: the pin **stays** once set, including when the model is cleared again. Nothing distinguishes an implicit pin from an explicit one, and the `<select>` names the connection from that moment on, so the route on screen is the route that will run and Default is one click away. My first draft of that comment claimed the pin was released; the comment and its test are corrected rather than left describing behaviour that doesn't exist.

Rebased onto `main` (`b67ef65`) earlier in this sequence; head is now `5a3cdd6`.

@codex review

---

charlesmsiegel:
Fixed in `547271d`. Third round running, third finding inside the previous round's fix — the observation is fair and the window is real: the pin reads its connection from an async `/api/config`, and the box was editable from the first render, so a model typed before the answer arrived stored `connection_id: ""` and was never revisited.

The control now refuses input until it knows the route, which is more than the narrow fix: the box's placeholder *is* the model its route will run, so until the route is known it has nothing true to tell the reader and nothing to attribute what they type to. `ModelCombobox` grows a `disabled` prop for it, which also suppresses its dropdown.

"Settled" is tracked separately from the value, because `active === null` is two different states — *not asked yet* and *asked, nothing is active* — and only the first should refuse. A read that **fails** settles too: we then genuinely don't know the active connection, and disabling the control forever over that is worse than letting an unpinned model through. An explicitly named connection is its own pin and never waits on the read.

Three tests cover it: the pending window, the failed read, and the explicitly-named connection. Worth noting one correction while writing them — `fireEvent.change` dispatches synthetically and ignores `disabled` (unlike `fireEvent.click`, which RTL gates), so "the spy was not called" proves nothing and `toBeDisabled` is the assertion that means what the test says.

@codex review

---

charlesmsiegel:
Fixed in `5b57e73`. Correct catch: `ModelCombobox` already stops Enter while its list is up for exactly this reason, and a native `<select>` has nothing equivalent to hook, so the guard belongs beside it. Escape is deliberately left to bubble — backing out of the popover from any of its three controls is what an earlier round added on purpose, and only Enter has a native meaning here worth protecting. Covered by `Enter on the connection select belongs to the menu, not to the popover`, which also pins that Escape still propagates.

Unrelated note for anyone reading CI on this PR: `CampaignView.test.tsx > a player post says what answering it cost, over every reroll` flakes intermittently under a full parallel run. I checked it against unmodified `origin/main` in a separate worktree — clean twice, then the identical failure on the third run with none of this branch's code present. It is the settle-timing class described in #351, not something this PR introduced, and the test never mounts the reroll picker at all.

@codex review

---

Issue #247, opened 2026-07-28
Overlay materialization writes record and manifest non-atomically


charlesmsiegel:
Surfaced by the codebase atlas (`docs/codemap.html`, generated at 81b61ff1). Same durability family as #233.

## Problem

Materializing an inherited record into a campaign is a two-step write with no atomicity between the steps:

```python
# backend/src/grimoire/store/overlay.py:73-90
dst.write_text(src.read_text(encoding="utf-8"), encoding="utf-8")   # step 1
manifest = campaigns.read_manifest(cid)
manifest[_flat_ref(kind, eid)] = entities.entity_hash(wroot, kind, eid) or ""
campaigns.write_manifest(cid, manifest)                              # step 2
```

A crash between them leaves a materialized record with **no recorded sync base**. The three-hash comparison then classifies it differently than it should:

```python
# backend/src/grimoire/store/sync.py:97-100
if world_h is None or world_h == rec["base"]:
    continue
status = "update" if mine_h == rec["base"] else "conflict"
```

A missing base makes an unchanged inherited copy look like a user edit — so a later world change reports as a **conflict** requiring manual resolution, rather than a clean update.

`materialize_actor` has the same shape but wider (`store/overlay.py:289-308`): it copies an entire directory of version files before writing the manifest, so the window is proportional to the actor's version count.

## Suggested fix

Write the manifest entry first (it is recoverable/idempotent), or stage the copy and commit both under one atomic rename. Pairs naturally with the shared atomic-write helper proposed in #233.

Issue #225, opened 2026-07-12
World-side delete + same-name recreate re-attaches campaign play/group state


charlesmsiegel:
Follow-up from the entity-kinds final review (branch entity-kinds).

Campaign-side deletion now cleans up group state (overlay.delete_entity unlinks groups/<gid>/state.md — see the entity-kinds branch). But a WORLD-route delete of a group or character followed by a same-name world recreate re-attaches any dependent campaign's orphaned state file (groups/<gid>/state.md, characters/<cid>/state.md) to the unrelated new record, feeding stale Knows/Secrets into scene context. This is pre-existing behavior for characters (playstate) and now shared by groups — fix both together, e.g. by sweeping dependent campaigns on world delete (cf. the demote/dependents design in #52) or by stamping state files with a record-creation nonce.

Related: #47, #52.

Issue #113, opened 2026-07-11
Show which campaign records have drifted from the library


charlesmsiegel:
Part of the **Extraction, Continuity & State** milestone. From the pre-rebuild target feature list:

> Library-drift → campaign-local override (never a library edit)

## Current state
This is **largely already satisfied by the copy-on-create architecture**. A campaign is a deep copy of its world (`store/campaigns.py:create_campaign`), and every write path that touches a campaign's records targets the campaign root exclusively: `absorb.apply_edits` (`backend/src/grimoire/store/absorb.py`) resolves `croot = campaigns.campaign_root(cid)` and calls `playstate.write_state(croot, ...)`, `entities.update_entity(croot, ...)`, `characters.update_version(croot, ...)`, `relationships.set_feeling(cid, ...)` — none of these ever touch a `worlds/<wid>/...` path. Manual edits through `EntityEditor.tsx`/`CharacterEditor.tsx` at campaign scope go through the same `update_entity`/`update_version` functions against the campaign root. So "drift never reaches the library" already holds by construction — there is no code path (play-driven or manual) that writes to a world file from campaign play.

What's missing is the second half of the feature: **explicit provenance** — knowing, for a given campaign record, whether it currently diverges from the world because of play/manual edits, versus because a world sync was accepted. The raw material already exists but isn't connected: `sync.md` (`campaigns.read_manifest`) stores the base hash `sync.py` last synced each ref against; `changes.json` (`store/changes.py`) records the latest write-back delta (scene + before/after) for browsable kinds (`character_state`/`lore`/`authored`). Nothing today compares "current entity hash vs `sync.md` base" to flag a record as drifted, and `changes.json` doesn't cover relationship/bond/plot edits at all.

### Option A — derived "drifted" badge (recommended)
- Compute `entities.entity_hash(croot, kind, eid) != sync.md base` (or `characters.dir_hash`/`card_hash` for actors) at read time; when true and a `changes.json` entry exists for that ref, badge the record "diverged via play" in `EntityEditor`/`CharacterEditor`'s sidebar.
- No schema change, reuses `sync.py`'s existing three-hash comparison logic and `changes.py`'s existing log — purely additive, deterministic.

### Option B — provenance field in changes.json
- Extend `changes.record` (`store/changes.py`) to tag each entry `source: "absorb" | "manual"`, distinguishing scene-driven writes from direct editor saves. Requires threading a source flag through every caller of `changes.record`/`apply_edits`.

[...]

PR #420, opened 2026-08-22
Records can move the other way: promote, push, demote (#52, #53, #60)

### LINKED ISSUE #52: Promote a campaign-local entity into the library, and demote it back
Part of the **Worlds & Library Content** milestone. From the pre-rebuild target feature list:

> Promote campaign-local entity → library; demote (reverse) with dependent-campaign warnings + optional copy-down

## Current state
Sync is strictly one-way world→campaign: `backend/src/grimoire/store/sync.py` diffs three hashes per ref (world / `sync.md` base via `campaigns.read_manifest` / campaign copy) and `accept`/`reject` only ever advance the campaign. Entities created campaign-side (`POST /campaigns/{cid}/{kind}`) get **no** `sync.md` base ref, and `sync.incoming` skips refs with no world hash — so campaign-local entities are permanently invisible to sync. There is no promote path.

On the demote side, deleting a world entity is silent by design: `incoming` skips world-side deletions (`world_h is None → continue`), so dependent campaigns keep their copies with no warning and no copy-down bookkeeping.

### Option A — Dedicated promote/demote routes (recommended)
- **Promote**: `POST /campaigns/{cid}/{kind}/{eid}/promote` — copy the campaign file (and `<kind>/<eid>/assets/` if present, cf. `campaigns.create_campaign`) into `worlds.world_root(campaign.md's world)`; on id collision return 409 with the existing record so the user chooses overwrite/rename. Then set `manifest[f"{kind}/{eid}"] = entities.entity_hash(wroot, kind, eid)` computed **from the newly written world file** — since the copies are byte-identical, the promoting campaign sees no incoming item; every other campaign of that world sees a normal `"new"` via sync.
- **Demote**: `GET /worlds/{wid}/{kind}/{eid}/dependents` returns campaigns whose `sync.md` contains the ref (sweep `campaigns.list_campaigns` filtered by world, `read_manifest`); `POST /worlds/{wid}/{kind}/{eid}/demote` with `{copy_down: bool, target?: cid}` copies the world file into any dependent campaign missing a copy, drops the ref from each manifest, then `entities.delete_entity(wroot, ...)`.
- Limit v1 to flat `entities.SYNCED_KINDS` (`locations`, `lore`, `greetings`).
- Recommended: keeps the base-hash bookkeeping server-side where it can't be skipped.

### Option B — "Outgoing" section in the sync review
- Extend `GET /campaigns/{cid}/incoming` (or add `/outgoing`) listing campaign-only entities with promote actions; UI rides the existing sync panel.
- One surface for all world↔campaign movement, but overloads the "incoming changes" mental model.

### Option C — Frontend-only copy affordance
- "Copy to world" button in `EntityEditor.tsx` calling the existing world create endpoint.
- **Broken by design**: without writing the base hash, the promoting campaign's next sync sees world-hash ≠ base(None) with mine ≠ base → a false `conflict`. The base-hash write is why this needs a backend route.

## Implementation notes
- Base-hash discipline is the whole trick: after promote, `mine == world == base` must hold byte-for-byte (write the world file as an exact copy of the campaign file; hash via `entities.entity_hash`, which is sha256 of full text through `store/statcache.py`).
- Ids are stable for life (`entities.py` docstring) — never re-slug on promote/demote; sync refs `(kind, id)` must keep lining up.
- Actors (characters/pcs) use `dir_hash` + `appearances` version-locking — out of scope here; #60 covers emergent-character promotion.
- Tests: round-trip in `backend/tests/` — create campaign entity, promote, assert `sync.incoming(cid) == []` for the promoter and `"new"` for a sibling campaign; demote with copy-down, assert file exists campaign-side and manifest ref removed.

## Dependencies & related
- #53 — pushing an *override* of an existing library record back (sibling operation; same base-hash mechanics).
- #60 — promote emergent characters (actor-shaped variant).
- #48 — emergent generated locations are prime promote candidates.
- #124 — promote an "extra" up to library scope.

### LINKED ISSUE #53: Save a campaign override back into the library as a new version
Part of the **Worlds & Library Content** milestone. From the pre-rebuild target feature list:

> Save a campaign override back into the library (bump version, clear the override)

## Current state
Campaign copies diverge freely from the world, and `backend/src/grimoire/store/sync.py` only detects that divergence when *pulling*: a campaign edit alone (mine ≠ base, world == base) produces no incoming item and nothing ever pushes it back. "The override" here is simply a campaign file whose hash differs from its `sync.md` base (`campaigns.read_manifest`/`write_manifest`); "clearing" it means making world and campaign byte-identical and advancing the base.

Note on "bump version": flat entities (`locations`/`lore`/`greetings`) have **no version field** — only content hashes. Real versions exist for actors: characters keep one V3 card per `<vid>.json` (`store/characters.py`: `create_version`, `set_default_version`), campaigns lock a version via `store/appearances.py` (`locked_version`, `set_base`), and world→campaign updates of locked actors flow through `POST /campaigns/{cid}/{kind}/{aid}/import-version`. The reverse (campaign→world) exists for neither.

### Option A — Entity push-back route (recommended)
- `POST /campaigns/{cid}/{kind}/{eid}/push` for `entities.SYNCED_KINDS`: write the campaign file's exact content to the world file, then set `manifest[ref]` to the world hash recomputed after the write — world == campaign == base, so the override is cleared and `sync.incoming` for this campaign is empty; sibling campaigns receive it as a normal `update`.
- Guard: if the world moved since base (`entity_hash(wroot,…) != base`), that's a *push conflict* — return 409 and require an explicit force (mirror of sync's pull conflict).
- UI: a "Save to library" button in the campaign-scope `EntityEditor.tsx` detail sidebar, shown only when diverged.

### Option B — Actor version bump
- For a locked character: `POST /campaigns/{cid}/characters/{aid}/push-version` creates a **new world version** from the campaign card (`characters.create_version(wroot, …)` — this is the literal "bump version"), optionally `set_default_version`; then `appearances.set_base(cid, kind, aid, actor_hash)` so no false pending appears.
- Preserves world history (old version untouched); complements Option A rather than replacing it.

### Option C — Outgoing-changes review panel
- A campaign-level view of every diverged ref (`mine != base` sweep, the inverse of `sync.incoming`) with per-ref diff preview (`changes.line_diff` in `store/changes.py` gives tagged op lines) and push buttons.
- Bulk-friendly superset of A; more UI and a new endpoint (`GET /campaigns/{cid}/outgoing`).

## Implementation notes
- False-conflict avoidance is the crux: compute the post-push base from the **world file after writing** (sha256 of full text via `entities.entity_hash` / `store/statcache.py`), not from the campaign file, in case any write path normalizes frontmatter.
- Never push automatically — this deliberately relaxes the "library-drift becomes a campaign-local override, never a library edit" invariant (#113), so it must be an explicit user action.
- Consider bumping `world.md`'s `updated` stamp — today only `worlds.rename_world` touches it; a library edit should probably refresh it (feeds #55).
- Tests: round-trip in `backend/tests/` under `GRIMOIRE_HOME` tmp_path — edit campaign copy, push, assert world content updated, `sync.incoming(cid) == []`, and a sibling campaign sees `"update"`.

## Dependencies & related
- #52 — promote/demote of campaign-*created* entities (this issue covers campaign-*edited* copies of existing library records).
- the sync-engine parent tracking issue (not re-filed as its own issue) — parent tracking issue for the sync engine this extends.
- #113 — the invariant this feature intentionally overrides on explicit request.

### LINKED ISSUE #60: Characters that emerge in play, campaign-local and promotable to the library
Part of the **Characters** milestone. From the pre-rebuild target feature list:

> Emergent (campaign-local) characters + promote-to-library

## Current state

Campaign characters today are always copies: deep-copied from a world at campaign creation (`campaigns.create_campaign`, `store/campaigns.py`) or synced in afterward via `store/sync.py`, which is strictly **world→campaign**: `incoming`/`accept`/`reject` diff three hashes (world/base/mine) per ref and there is no reverse direction anywhere in that module.

There is no route to create a brand-new character that lives only in a campaign. The campaign character routes (`routes.py:1631-1688`) cover list/get/set-default-version/create-version/update-version/delete-version on an **existing** campaign character folder — there is no `POST /campaigns/{cid}/characters` (contrast the world side, which has a create route). The store function itself, `characters.create_character(root, name, ...)` (`store/characters.py:78`), is root-agnostic — it would work unmodified pointed at a campaign root — so the store layer isn't the blocker, only the route/UI wiring. There is also no promote-to-library (campaign→world) path anywhere; `sync.py`'s three-hash diff only ever flows world changes into a campaign, never the reverse.

### Option A — campaign-scoped create route + explicit promote action (recommended)
- Add `POST /campaigns/{cid}/characters` calling `characters.create_character(campaign_root, ...)` directly, with no world-side counterpart. The character being absent from `sync.md`/the world's `characters/` folder IS the "emergent" signal — no new field required.
- Add `POST /campaigns/{cid}/characters/{cid}/promote`: deep-copy the campaign folder into the corresponding world's `characters/` (mirroring, in reverse, the copy logic `create_campaign` already performs), then record a `sync.md` base hash so the campaign copy becomes a normal synced/overridable actor going forward.
- Frontend: `CastPanel.tsx` gets a "New NPC" action (campaign-only) and a "Promote to library" button on an unpromoted cast member.
- Tradeoff: reuses all existing character-store plumbing; the only new code is the two routes plus the promote copy step.

### Option B — implicit emergent character via unknown-name routing
- Pair with #98 (unknown-name routing): when prose introduces an unrecognized name, offer to auto-create it as a campaign-local character via Option A's create route.
- More automated end-to-end, but depends on #98 landing first — better framed as a follow-on than the base of this issue.

## Implementation notes
- Library content is world-scoped; a promoted character must become indistinguishable from one authored directly in the world — reuse `create_campaign`'s copy mechanics rather than a bespoke path.
- Campaign-local-only characters (never promoted) must be excluded from `entities.synced_refs`/`sync.incoming` so they never surface as a phantom "world change."
- Tests: backend pytest, `monkeypatch.setenv("GRIMOIRE_HOME", tmp_path)`; cover create-without-world-counterpart and promote-then-normal-sync-divergence.

## Dependencies & related
- #52 — the analogous promote-to-library flow for entities (lore/locations); this issue is the character-kind counterpart.
- #61, #62 — natural follow-ons operating on an emergent character once it exists.
- #98 — unknown-name routing, a possible producer of emergent characters (Option B).

charlesmsiegel:
Closes #52, closes #53, closes #60.

Campaign↔world movement was one-way. `sync.incoming`/`accept`/`reject` compare three hashes per ref (world / base / mine) and only ever advance the campaign, so a location invented at the table, an NPC who walked on mid-scene, and a campaign's better wording of a library entry all stayed where they were written.

This adds the reverse direction as three explicit, user-driven moves, plus the one missing create route. Nothing here ever runs on its own: all three relax the rule that library drift becomes a campaign-local override and never a library edit (#113), so each is something a user asks for by name.

## What the overlay changed about the question

All three issues predate the copy-on-write overlay, and it moves two of their premises — `docs/superpowers/specs/2026-08-21-library-promote-demote-design.md` carries the reasoning:

- **A promoted record reaches sibling campaigns live**, not as a sync `"new"`. A campaign materializes a record only when it diverges; everything else resolves through to the world. (`sync.incoming` never emits `"new"` at all — the key in `campaigns_for_world`'s counter is vestigial.)
- **"Dependent campaigns" means every campaign of the world**, not only the ones holding a manifest ref. Under the overlay, a campaign with no copy is the one depending on the world record *most* — it has nothing else.

Demote is therefore built on `overlay.forget_world_record`, whose docstring already named #52 as the issue carrying the dependents design.

## The moves

| | precondition | effect |
|---|---|---|
| **promote** (#52, #60) | the world holds no record under that id | copies the campaign record up and records the base, so `mine == world == base` |
| **push** (#53) | the world holds it and the campaign has a diverged copy | the world takes the campaign's bytes, the base advances, the override clears |
| **demote** (#52) | the world holds it | optionally copies down to every dependent, then deletes and sweeps |

`POST /campaigns/{cid}/characters` is the create route the campaign side never had — the world had it all along. A character with no world counterpart and no sync ref *is* the "emergent" signal (#60); nothing enumerates it, so it can never surface as a phantom world change.

## Base-hash discipline, and which way each move fails

The two writes each move makes cannot be made one, so each fails onto the harmless side — the argument `overlay._recorded_base` makes for #247, applied in both directions.

- **Promote writes the base first.** Its residue is a base whose world record never landed: sync skips it, the record still reads campaign-side, and a retry completes it. That is also why promote is gated on the **world's** state rather than on the manifest ref — a ref-based gate would refuse the very state the crash left behind.
- **Push writes the world record first.** Its residue is world and campaign agreeing with a stale base. A retry heals that (advances the base and clears the override) instead of reporting a conflict between two identical files.
- **Demote copies down before deleting**, and aborts if any copy-down fails — a partial copy-down is the state before the call; deleting first would lose the record for every campaign the copy never reached.

## Refusals

Typed (`LibraryMoveError` and friends), mapped to 409 with a `kind` the UI branches on. `push_conflict` is the only one the user can resolve, so it is the only one that gets an overwrite button. `libraryStatus` decides which action an editor offers **server-side**, so a second copy of the rule cannot drift into offering the button that 409s.

## Scope, stated rather than quietly dropped

- **Push does not carry actors.** A locked actor's base lives in `appearances.json` and pushing one means minting a *new world version* (#53's Option B) rather than overwriting the current one — a different operation, refused with a message that says so.
- **Demote is flat kinds only**, per #52's own v1 boundary.
- **Plot-map edges do not travel** with a promoted greeting, and a demoted greeting's edges leave the world's map as an ordinary world-side delete already does.
- **`GET /campaigns/{cid}/diverged`** (#53's Option C read half) is API-only — tested, but no UI consumes it yet. The per-record button is Option A, which is what the issue recommends.

## Review findings, each reproduced before it was fixed

Five adversarial passes. The three destructive ones:

- **A promote into a deleted world rebuilt it.** `world_root_of` answers with a path whether or not a world is there, and `mkdir(parents=True)` recreated the directory around the promoted file — a tree with no `world.md`, which nothing lists as a world and no route can reach, holding the only copy of a record the campaign had just recorded a base for.
- **A demote to an unknown target deleted the record everywhere.** The delete ran regardless of what `target` matched, so a typo meant "copy this down nowhere, then take it away from every campaign".
- **A demote destroyed the record's art.** Assets overlay per *file* from the world, so copying only the text down left every dependent holding a record whose pictures were gone for good. `overlay.copy_record_dir_down` carries them, obeying the two overlay rules that make it safe: a file the campaign already has is never overwritten, and a tombstoned asset stays gone.

Plus: ids are refused before they reach a world path (everything here writes to the world, which no campaign resolver guards); a greeting whose character is *detached* is refused, because filing it against the stranger holding that slug reads as working (#225); asset copying can no longer fail a promotion that has already committed; and both asset copies go through `atomic.write_bytes` rather than `shutil`, per the convention `sheets/tally.py` states at the identical copy — a torn image in a campaign's record directory would *shadow* the world's intact one, since the overlay resolves images campaign-first.

## Testing

- `backend/tests/test_library_move.py` — 50 store-level cases: round trips, crash residue, detachment, assets, referential integrity, traversal.
- `backend/tests/test_library_routes.py` — 29 route cases: status codes, the refusal `kind`s, the full promote → library edit → accept loop.
- `frontend/src/components/LibraryPanel.test.tsx` (11) and `DemotePanel.test.tsx` (10) — both new components at 100% line coverage.
- `test_route_order.py` caught a real crossing (`/sheets/{kind}/{eid}` vs `/{kind}/{eid}/library`); it is pinned with its winner and the reason.

Backend 6627 passed, frontend 1914 passed, ruff/mypy/eslint all at baseline, templates byte-identical, and green again under the Android dependency set (pydantic 1.10).

---
_Generated by [Claude Code](https://claude.ai/code/session_018DwZtcPrf9w3SUXBeguSMD)_


charlesmsiegel:
Round one addressed in 98905d7 — seven of the eight findings were real and are fixed, each with a regression test:

- **P1, demote held no campaign locks** — now `locks.hold_all` over the *full* dependent set (not the `target`-narrowed one, since the sweep touches all of them).
- **Actor promotion could not retry** — `_world_holds_actor` asked `is_dir()`, but the directory precedes the version files and the meta lands last, so an interrupted attempt was refused as a collision forever. It now keys on the meta file (what `overlay.actor_root` and `_record_present` already treat as the commit point) and purges the residue, as `materialize_actor` does campaign-side. This was the exact stuck state the base-first ordering exists to prevent — built for flat records, missed for actors.
- **A failed promotion left a detached record attached** — `_reattaching_on_failure` restores it on the exception path.
- **A greeting's `present` cast went unchecked** — same existence and detachment checks now apply to every referenced character.
- **`touch` could revert a rename** — it re-reads and skips rather than forcing; it is only a sort key, and there is no world lock to hold.
- **A demoted greeting left a pre-filled draft** — `resetForm()`, matching the entity editor.
- **Stale dependents with a live button** — the panel now says it is checking and withholds the destructive action until the refresh settles.
- **PC promotion had no UI** — `PCEditor` mounts the panel.

One I did not take: the claim that `Run`, `Other`, `The Long Run` and `Saltmarch Nights` are newly invented fixture names. All four are already established in this repo's tests (`test_sync_store.py`, `test_search_store.py`, `test_covers_store.py`, `test_campaign_images_store.py`, `test_semsearch_store.py`), and CLAUDE.md asks for exactly that — reuse over invention. `Started Since` *was* mine and has been replaced.

@codex review

---

charlesmsiegel:
Round two addressed in 6e374f3 — all ten findings were real. The three that lose or corrupt data:

- **Actor promotion adopted the dead actor's face.** Round one's purge covered only root-level version files, so `assets/`, `tagline.md` and `voice_anchor.md` survived — and those are overwritten only where the campaign supplies one, so an empty slot kept the previous actor-of-that-slug's picture and voice. The whole meta-less directory goes now.
- **Promote and push republished `sync.md` unserialized.** Both now hold `campaign_lock(cid)` across the read-modify-write, the same lock demote takes.
- **A demoted greeting lost its edges.** Copy-down materializes each dependent's plot map before the world delete, so the record it was just handed keeps its prerequisites.

Plus: `undetach` moved inside the restore handler's `try` (it is two writes); push now runs promote's referential checks on greetings; the greeting's `version` is validated against the world actor, not just the actor; `library_status` compares actor hashes so a rewritten character is no longer reported "in sync" (it is diverged, it simply cannot be pushed — the panel says that now); the push heal path stamps the world; "Remove everywhere" is gone as a promise demote does not keep; and a failed dependents refresh leaves Try again / Cancel instead of "checking…" forever.

Each has a regression test. Backend 6634 passed, frontend 1918 passed, ruff/mypy/eslint at baseline.

@codex review

---

charlesmsiegel:
Round three addressed in 57fc072 — all six were real, and the P1 was the headline case.

- **P1, promoting an NPC who had already walked on.** Casting an emergent actor locks a version in `appearances.json` with an *empty* world base, and `_actor_incoming` prefers that record to `sync.md` — so the promotion succeeded and then immediately offered itself back as a conflict to the campaign that made it. Promotion now advances the appearance base to the world version it just wrote. Two tests: nothing incoming after promoting a cast NPC, and a later library edit still arriving as an ordinary `update` (so the advanced base is the *right* hash, not merely non-empty).
- **Non-UTF-8 sidecar** — `_copy_extras` now catches `ValueError` too, so a committed promotion is not reported as a failure whose retry then collides.
- **`present: mara, winifred`** — validated as it will be *read*, since `_tags_list` does not strip.
- **Push used `_world_holds_flat`** — directory occupancy is right for blocking a promote and wrong for push; it asks for the record file now.
- **Concurrent cross-campaign push** — the conflict check is re-taken immediately before the write, narrowing it. Closing it fully needs a world-scoped lock, and this store has none for anything (world create/rename/delete are all unserialized), so introducing one for a single call site would add a second lock domain and a new ordering pair to deadlock on. Written down in the design doc as a known gap rather than papered over.
- **No way to delete a campaign-created NPC** — correct, and mine: `overlay.delete_actor` is the actor twin of `delete_entity`, with the same three cases, and it drops the appearance record too.

That last one surfaced something worth flagging: `test_clearing_a_flag_on_a_deleted_character_still_works` built "a deleted character" with a `DELETE /campaigns/{cid}/characters/{char}` that **silently 404'd**, because no such route existed and the request fell through to the generic entity handler. It was really testing a character that still existed. I rewrote it to build the state it describes — a flag filed under an id no character claims — and added a separate test for the behaviour it gave up: a real delete takes the flag with it, which is `delete_entity`'s rule for sidecars and #225's reason for it.

Backend 6645 passed, frontend 1919 passed, ruff/mypy/eslint at baseline.

@codex review

---

PR #422, opened 2026-08-22
A catalog from the provider you configured, and a status that means it works (#146, #149)

### LINKED ISSUE #146: Provider health checks, on demand and periodic
Part of the **LLM / Providers** milestone. From the pre-rebuild target feature list:

> Provider health checks (on-demand + periodic) with status

## Current state

**⚠️ Pre-rebuild concept — needs partial re-scoping.** "Periodic" implies a background scheduler; nothing async-periodic exists anywhere in this codebase (no cron/interval task, no APScheduler-style dependency in `backend/pyproject.toml`, no lifespan background loop beyond the one-shot `store.migrations.migrate_scene_ids()` run in `main.py:create_app()`'s lifespan hook). Building a periodic poller would be the first scheduled-task infrastructure in the app.

No health check of any kind exists today. There is no `GET /api/.../health`-style route, no ping helper in `openrouter.py`, and no status surfaced anywhere in the frontend (the topbar shows only "OpenRouter key status" — i.e. whether a key string is present in `config.md`, not whether it actually authenticates — per the architecture doc's summary of the topbar). The only signal a user currently gets that OpenRouter is unreachable/misconfigured is a failed chat/absorb/tagline call surfacing `OpenRouterError.kind` (`missing_key|auth|rate_limit|network|bad_response`) after the fact.

### Option A — On-demand ping only, no scheduler (recommended)
- New route, e.g. `POST /api/config/health-check` (or a query param on `GET /config`): fires a minimal request against OpenRouter — either `GET https://openrouter.ai/api/v1/models` (cheap, no completion cost, also validates the key can authenticate against OpenRouter's auth) or a 1-token `.complete()` call — and returns `{ok, kind?, detail?, checked_at}`.
- Frontend: a "Test connection" button in `ConfigView.tsx` next to the key field, showing the result inline; optionally also call it once at topbar mount so the existing "key set" indicator becomes a real "key works" indicator.
- No new persistent state, no scheduler. Recommended: matches the "prefer on-demand + on-error" guidance — periodic polling has no infrastructure to build on and a chat failure already tells the user something's wrong in real time.

### Option B — On-demand + "on-error" passive status
- Same on-demand check, plus: every real `OpenRouterError` raised by any call site (`routes.py`'s nine call sites) updates an in-memory `last_error: {kind, detail, at}` alongside a `last_success: {at}` timestamp (module-level state next to `_openrouter` in `routes.py`, or a small `store/health.py` if it should persist across restarts).
- `GET /api/config` (or a new `/api/health`) includes this so the topbar can show "last error 2 min ago" without polling.
- Tradeoff: gives a genuinely useful "is it healthy right now" signal cheaply, still no scheduler.

### Option C — True periodic health check
- Add a background task via FastAPI's `lifespan` (an `asyncio.create_task` loop with an interval, cancelled on shutdown) that pings on an interval and updates the same status state as Option B.
- Tradeoff: real scheduled-task infrastructure for a single use case; over-engineered until something else in the app also needs periodic background work. Not recommended as a first step.

## Implementation notes
- Reuse `OpenRouterClient`'s existing error normalization (`OpenRouterError(kind)`) rather than inventing a parallel status vocabulary.
- `_require_key` (`routes.py`) already returns 409 `missing_key` before any network call when the key is empty — a health check should short-circuit the same way rather than making a doomed request.
- Tests: `backend/tests/test_routes.py`'s `app.dependency_overrides[routes.get_openrouter]` fake-client pattern applies directly — assert the health route reports `ok:false` with the right `kind` for a fake that raises, and `ok:true` for one that responds. Store isolation: `monkeypatch.setenv("GRIMOIRE_HOME", tmp_path)`.
- If Option A's `GET .../models` ping is chosen, note the frontend already fetches that exact URL directly from the browser (`frontend/src/api/models.ts:fetchModels`, unauthenticated) for the model catalog (#149) — a backend health check is a different call (server-side, can include the key) and shouldn't be confused with that one.

## Dependencies & related
- #141 — once a configurable base URL/provider exists, health checks should target the active provider, not a hardcoded OpenRouter URL.
- #147 — a Providers tab is the natural home for a visible health/status indicator.
- #154 — performance metrics (latency, error counts) could subsume or complement per-call health signal from Option B.

### LINKED ISSUE #149: Point the model picker at the configured provider, not always OpenRouter
Part of the **LLM / Providers** milestone. From the pre-rebuild target feature list:

> Model catalog picker from a configured provider

## Current state (largely implemented — parent for remaining gaps)

The core of this feature already exists and works: `frontend/src/routes/ModelCombobox.tsx` is a live-search combobox used by `ConfigView.tsx`'s Model field. It calls `getModels()` (`frontend/src/api/models.ts`), which fetches `https://openrouter.ai/api/v1/models` directly from the browser, maps each entry to `{id, name, context: context_length, prompt: pricing.prompt, completion: pricing.completion}`, sorts by id, and caches the result for the page session (`modelsCache`, invalidated via `invalidateModelsCache()`). The combobox shows the full catalog on focus and narrows by id/name/price text as the user types (`matches` filter in `ModelCombobox.tsx`); each row already renders **both** requested metadata fields — price (`priceLabel`: tokens-per-dollar or "Free") and context length (`contextLabel`: compact "K/M ctx") — so pricing/context-length metadata display is done, not a gap.

- [x] Fetches the live model catalog (not a hardcoded list) — `models.ts:fetchModels`
- [x] Search/filter combobox UI — `ModelCombobox.tsx`
- [x] Pricing metadata shown per model — `priceLabel`/`tokensPerDollar`
- [x] Context-length metadata shown per model — `contextLabel`
- [x] Also consumed elsewhere for display: `frontend/src/components/SceneInspector.tsx` uses `getModels()` to look up a scene's stamped model's context size

What's not done:
- [ ] "From a **configured** provider" — the catalog URL is hardcoded to `https://openrouter.ai/api/v1/models` in `models.ts`; there is no notion of a configurable base URL/provider yet (#141), so today the picker is really "OpenRouter's catalog, always" rather than "whichever provider is configured." Once #141 lands (whichever option), `fetchModels()` needs to target the active provider's models endpoint instead of the hardcoded constant — and local/OpenAI-compatible servers that don't expose a `/models` endpoint (or expose a different shape) need a graceful fallback to free-text entry, which `ModelCombobox` already does today on fetch failure (`error` state → "couldn't load model list — type a model id").
- [ ] Per-task pickers (#142) — `ModelCombobox` is reused as-is wherever a second/third model field is needed (e.g. a Light-tier or per-campaign override picker); no work is needed in this component itself, just additional call sites once #142/#143 exist.
- [ ] The fetch is unauthenticated and client-side (browser calls OpenRouter directly, no key attached) — fine for OpenRouter's public catalog endpoint, but a future local/custom provider's model-list endpoint might need the request to go through the backend (with the key) instead of straight from the browser; worth revisiting once #141 picks a provider shape.

## Implementation notes
- Frontend tests: vitest run from `frontend/` (`npx vitest run`) + `npx tsc -b`; `models.ts`/`ModelCombobox.tsx` should already have `.test.ts(x)` siblings per the project convention — extend those rather than adding a parallel test file when #141 makes the URL configurable.
- When #141 is implemented, the minimal change here is: replace the hardcoded `MODELS_URL` constant in `models.ts` with a value sourced from `GET /config` (or wherever #141 stores the base URL), keep the existing `Model` type/parsing as-is since it already matches OpenRouter's response shape.

## Dependencies & related
- #141 — provider/base-URL configurability; the only real remaining gap for "from a configured provider."
- #142 — per-task model routing; more pickers reusing this same component, no changes needed here.
- #147 — Providers tab would embed this picker inside the LLM provider card.

charlesmsiegel:
Closes #149. Closes #146.

Two issues that turn out to be one seam.

**#149's remaining gap** was that the model picker fetched
`https://openrouter.ai/api/v1/models` from the browser, unauthenticated,
whichever provider a connection actually named — so a reader on a local
endpoint picked from a catalog their connection could not run, and a rejected
key first surfaced as a failed scene.

**#146** asked for a health check, because nothing in the app had ever asked a
provider anything: the topbar's green dot meant "a key string is present in
`config.md`", and the Configuration page said "key set" in words.

Both are questions about one connection that only the facade can answer, so
`LLMClient` gains the two calls that dispatch by kind alongside `stream`:

```
list_models(conn)   the connection's own catalog, normalized (catalog.py)
check(conn)         can this connection serve, right now
```

Neither is retried and neither falls back. A catalog answered by a different
provider lists ids this connection cannot run, and "is it healthy" answered by
trying something else is not an answer. That folds the second seam away:
`get_openai_compatible_client` existed only because listing was a capability
one provider had, and it is now a capability the facade dispatches.

## The catalog (#149)

- OpenRouter grows `list_models`/`probe`; both clients share one entry shape,
  sorted by id, so one combobox no longer lists two ways depending on which
  connection is open.
- `POST /llm-connections/{id}/models/refresh` covers every listable kind,
  presenting the stored key. `POST /model-catalog` answers for a connection
  *described but not saved* — the setup wizard's whole job on that step, and the
  New-connection form's.
- `claude` stays refused: its models are SDK aliases with nothing to enumerate,
  which is why that kind has a fixed list rather than a picker.
- The frontend keeps its cache and its labels and loses the fetch. A saved
  OpenRouter connection still fills its picker on open (that is what the browser
  fetch did), quietly — a fetch nobody asked for degrades the field to free text
  rather than raising a banner. A custom endpoint keeps its explicit button: it
  can be a local server that is switched off, and stalling on merely *looking* at
  a connection is worse than an empty list.

## The status (#146)

- `health.ProviderHealth` on `app.state`, per app for the reason the rest of the
  per-app state is: a `TestClient` builds an app per test and recorded failures
  must not leak between them. In memory, because it describes *now*.
- Two writers. `POST /llm-connections/{id}/health` is the on-demand half —
  always 200 with the verdict in the body, because the provider failed and the
  request did not. The facade's observer is the passive half: every generation
  that settles reports its outcome, which keeps the answer current between
  clicks.
- **No poller, deliberately.** #146 sketched one (its Option C) and recommended
  against it, and the passive half is why: a scene turn is a better probe than a
  poll of a cheaper endpoint, and what a poll adds is freshness while nobody is
  playing. A provider failure announces on the config channel from both places
  failures arrive — the api client's rejection path and `CampaignView.fail`,
  where an SSE error frame lands — so the dot tracks real traffic rather than
  waiting for a navigation.
- The dot gets a third state; the Connections page gets the reason in words
  beside the key it is about, plus a **Test connection** button and a "failing"
  badge in the rail; the Configuration page stops saying "key set" as if that
  were the answer (which also stops it telling a `claude` connection it has no
  key — a thing that path has never needed).
- Editing or deleting a connection forgets its verdict: the failure on record
  belonged to the setting the reader just changed, and slugs are reusable.
- The wizard now checks what it just saved, so "Connected ✓" stops meaning
  "saved and made active". A refusal is a warning, never a gate.

## Notes for the reviewer

- **Branch.** The harness assigned `claude/issues-146-149-ptf7uz`, but that
  branch already carries 44 unmerged commits of unrelated work (the detached-runs
  feature, #394). Building on it would have made this PR 14k lines of somebody
  else's change, so this went to its own branch off `main`.
- **`PUT /config/data-dir` is unaffected**; the client-side "a provider failed"
  signal is gated on the LLM error kinds specifically, because that route also
  answers with a `kind`.
- **Coverage.** `origin/main` measures **92.95%** against a 93% floor on this
  container; this branch measures **93.00%**. Both shortfalls (and the one
  failing `test_atomic` case) are artifacts of running the suite as root, where
  a `0444` file is still writable — CI runs unprivileged.
- Everything else in the gate is green: `ruff`, `mypy` and `eslint` all at
  baseline (the eslint baseline shrank by 6 pairs, since deleting the browser
  fetch removed its `any` handling), 6607 backend tests and 1915 frontend tests
  passing.

https://claude.ai/code/session_01K6Wkjcpm9ffu6ZsQZSxAET

---
_Generated by [Claude Code](https://claude.ai/code/session_01K6Wkjcpm9ffu6ZsQZSxAET)_


charlesmsiegel:
Worked all ten Codex findings. Eight taken, two not — details below, all pushed in 9cfbf86.

**Taken**

- **Malformed catalog envelopes → 500** (`openrouter.py`, `openai_compatible.py`) — correct and the worst of the ten. `{"data": null}` from a local server with nothing loaded reached the normalizer as `None` and raised a `TypeError` out of the route; a bare array raised `AttributeError` off `.get`. Both clients now validate the envelope and answer `bad_response`, which the picker already renders as "couldn't load model list — type a model id". `{"data": []}` stays a legitimately empty catalog. The `openai_compatible` check sits *outside* the request funnel deliberately — inside it, `except Exception` would relabel "your endpoint sent nonsense" as `network`.
- **Health check unbounded when `llm_call_budget: 0`** — correct. That setting's `0` means "no ceiling at all", which is reasonable for a slow local model and not for a button, and the Claude probe has no transport timeout of its own. The route now takes `HEALTH_CHECK_CEILING`, which is not a setting.
- **Budget-cancelled generations never reached the registry** — correct, and the sharpest of the ten: the ceiling *cancels* the stream, so `_resilient` unwinds through `GeneratorExit` rather than its `except LLMError`, leaving a 504 under a green dot. Cancellation can't be read as failure from inside the facade (an SSE client disconnecting unwinds identically and is nobody's fault), so the holder of the ceiling reports it, via a new `LLMClient.note_outcome`.
- **Rev check missing on `setModels`** — correct, and my own comment promised the check I hadn't written. Both halves now compare id *and* rev.
- **Health outliving the settings it described** — correct, and better than the caveat I'd written for it. Verdicts are now filed under the connection's revision and read back by it, so an attempt that started before an edit can't install the old key's failure against the new one, and a recreated slug can't inherit a dead connection's. The facade is handed whole connection dicts and those already carry `rev`, so this cost a key rather than a mechanism.
- **`refreshConnectionModels` didn't invalidate the page-load cache** — correct; it announces on the config channel now.
- **Recovery was silent** — correct. `fail` announced and success didn't, so a Retry that worked left the dot red until an unrelated navigation. Success announces too, but only when a failure was announced first: otherwise every message in the game would carry a config read behind it.
- **Wizard couldn't use the key it had just been given** — correct. The automatic fetch stays keyless (that's what makes the model field usable on the step where the key is still being typed), and it now has the same explicit Fetch models button the editor has, which also gives custom endpoints a catalog before the connection is saved.

**Not taken**

- **Validate the configured model as part of health.** A local OpenAI-compatible server is entitled to serve a model its `/models` doesn't list — llama.cpp and LM Studio both do — so "not in the catalog" isn't "cannot generate", and gating `ok` on catalog membership would report working connections as broken. The check answers "can this connection reach its provider and authenticate", which is the question that has a cheap, honest answer; choosing a valid model id is the picker's job.

Each fix has a test, and the two guard-shaped ones (the rev check, the ceiling) were confirmed to fail with the guard removed.

---

charlesmsiegel:
## Codex round three — eight findings, all worked

Pushed as `e1a760d`.

**P1 — inline fakes.** Four doubles had grown up inline across the new tests, against a rule this repo states plainly (`AGENTS.md`: LLM behaviour comes from `backend/tests/llm_fakes.py`, never a new inline fake). `StallingGateway`, `ScriptedProvider`, `FlakyProvider` and `RecordingProvider` now live in `llm_fakes.py`, and `FakeLLM` grew the surface the health work actually needs — `models` / `models_error` / `health_error` to script an outcome, and `listed` / `checked` / `noted` to read one back.

**P2s:**

- **A ceiling that fires after a failover was blaming the primary.** Two wrongs in one: it overwrote the primary's real failure with a timeout it did not cause, and it left the connection that actually stalled looking healthy. The facade already stamps the attempted connection into the usage holder (`llm.ATTEMPTED`), so `_noting` reads that and falls back to the route's own connection only for a caller that threads no holder.
- **`_noting` is now guarded**, and a red suite earned the comment above it: this runs on the failure path, so an exception inside it does not merely lose a status update — it *replaces* the error the caller is about to raise. A gateway double without `note_outcome` turned "the budget stopped this phase" into an `AttributeError`, and the phase stopped reporting why it died. Same reasoning `llm._observe` already carries.
- **`_Budget.run` calls `on_timeout` on the `TimeoutError` path only.** A `BudgetRefused` is the budget declining to *start* a call; there is no provider to blame for it.
- **`catalog.entry` asks whether `pricing` is a dict** instead of leaning on `or {}` — a row with `"pricing": "free"` normalizes now rather than crashing.
- **The verify skill's three-URL note had the mapping backwards** (`MODELS_URL` is the catalog, `KEY_URL` is the 200/401).

**Verification:** full backend suite 6623 passed; the three ratchets (ruff 1183, mypy 246, eslint 932) all at baseline; frontend suite green. The one remaining local failure is `test_atomic.py::test_a_read_only_record_is_not_silently_replaced`, which is a root-user artifact of this container — `chmod 0444` does not stop uid 0 — and it fails identically on `origin/main`. CI runs non-root.

---

charlesmsiegel:
@codex review

---

Issue #423, opened 2026-08-24
Off-scene cast directory loses its heading when the layout disables or reorders a half


charlesmsiegel:
`off_scene_cast_known.j2` decides who emits the shared heading from **data**, not from what actually rendered:

```jinja
{%- if not offscene_active -%}{%- set _ = parts.append(heading()) -%}{%- endif -%}
```

`offscene_active` is the tier-2 *data* (`assemble.py:221`), while `layout.py` explicitly lets a user disable, reorder and relabel any section. The two can disagree, and the macro's own docstring assumes they can't — it says the heading rides on "whichever renders first, which is tier 2 when it has content and tier 3 otherwise", conflating *has content* with *renders*.

### Reproduction

Disable `off_scene_cast_active` in the prompt layout while its data is non-empty. Tier 3 still sees `offscene_active` and suppresses the heading:

```
## Known to exist
Winifred: Keeps the ledger.
```

Expected — the heading tier 2 would have carried:

```
# Other characters in this world

# (Not present. Introduce them only if the story calls for it.)

## Known to exist
Winifred: Keeps the ledger.
```

### Why it matters beyond formatting

The lost heading is the line that says the cast is **not present**. What survives is a bare list of names under `## Known to exist` with nothing marking them absent — so the section most likely to make the model write an off-scene character into the scene is exactly the one that loses its guard. `assemble.py`'s own comment already flags the unframed-list case as the bad one; this is a second, layout-driven route into it that the comment doesn't cover.

### Also affected: reordering

Put `off_scene_cast_known` before `off_scene_cast_active` in the layout and the heading renders in the *middle* of the block — tier 3 suppresses it, tier 2 emits it second.

### Note

Any data-derived "am I first?" test has this shape, because the question is about the merged, enabled layout and the data cannot see it. `docs/superpowers/specs/2026-08-24-character-voice-in-scene-prompt-design.md` §1 declines to copy the pattern for this reason and gives its sections independent headings instead — that may or may not be the right fix here, since this directory genuinely is one block split in two for token accounting.

PR #424, opened 2026-08-25
Character voice reaches the scene prompt, not just the judge

charlesmsiegel:
Two NPCs in a scene produce lines that could be swapped without anyone noticing. The prompt fights *robotic* prose at length in `natural_prose.j2`; it fought *same-y* prose nowhere at all.

## What was wrong

The voice anchor was judge-side only. `context/cast.py` read it to validate a standing drift flag, and there was no `Section(...)` for it in the catalog — so it could flag same-y speech after a scene was played, and never prevent it during one. Nothing in the app ever wrote one either, so the whole subsystem was off by default rather than by choice.

Meanwhile `character_descriptions.j2` and `message_examples.j2` concatenated every present NPC's fields with **no name attached**. For SillyTavern-style cards, whose examples lean on `{{char}}` and unattributed narration, there was genuinely nothing tying a sample to the character it came from.

## What this does

**Three sections replace one**, immediately after `character_descriptions`:

| section | tier | why |
|---|---|---|
| `voice_policy` | `LOCK_IN` | fixed instruction text — the differentiation rule and precedence order, bounded by construction |
| `voice_anchors` | `SPOTLIGHT` | per-character information, cast-sized, so exactly what must not be pinned |
| `voice_examples` | `SPOTLIGHT` | same |

Anchors are **not** guaranteed to outlive examples — `pack` drops the largest *actual* section in a tier and a pinned one never drops — so nothing depends on that ordering and the test asserts it only for a case that forces it.

One `cast_blocks` structure feeds all four affected sections, so name attribution and nameless handling are defined once and cannot diverge between them.

**Each section carries its own heading.** This deliberately departs from `off_scene_cast_active/known`, which share one via a macro that picks its owner from *data* while `layout.py` lets a reader disable and reorder any section — #423.

**The drift judge is given the character's outstanding correction.** The policy tells the generator a correction outranks the anchor, so without this the judge flagged the model for obeying, and the flag minted another correction. `system.j2`, `user.j2` and `build_prompt` move together; changing only the user message builds an evaluator whose halves disagree. The caller validates provenance, so a note fingerprinted to a replaced anchor never reaches the judge.

**A retired section stays retired.** `layout.py` remaps a stored `message_examples` entry to `voice_examples`, keeping its position and `enabled: false` — ahead of `_ordered`, which `merge` and `describe` share, or the renderer would honour the disable while the editor reversed it on save.

## Judged, not proven

The anchor now steers generation, which spends a real property: the drift check no longer compares against an independent reference. That was taken deliberately, and the docstring says so — it is an **approximate second opinion**, comparing against the character's *current* anchor, which may differ from what a turn received (substitution, a packer drop, a layout disable, an edit between play and absorb, a local template edit). A `drift` verdict means "worth looking at". Making it exact would mean snapshotting the delivered brief with every accepted turn; that is priced in the spec and not taken.

What *is* held exactly: generator and judge receive the same `effective()` text, so a rule past the cap is enforced against neither.

## Notes for review

- `check-mypy` fails on `atomic.py` / `proclock.py` — files this branch never touches. Reproduces with everything stashed; a Windows-only artifact CI does not see. Deliberately **not** baselined, since that would record a Windows-specific count and break the gate on Linux.
- The frozen fixture's card gained a `mes_example`. Without one the voice sections rendered nothing against it and the snapshot showed no diff — indistinguishable from a feature that silently does not work. A fidelity edit, not coverage: a card of that vintage essentially always carries examples, and Mara's card is still empty so both cases stay covered. No `voice_anchor.md` was added; no store that old has one.

Spec: `docs/superpowers/specs/2026-08-24-character-voice-in-scene-prompt-design.md`

https://claude.ai/code/session_01Sdk4bPEMaNK7yhq3VfyLfN


charlesmsiegel:
Round 1 addressed in `f661e4e38`. Four of five taken.

**Disambiguation withdrawn** (P2, aliases vs speaker parsing) — you were right that this reaches further than the prompt, and it's worse than stated. `match_name("Winifred", ["Winifred","Winifred"])` is already `None`, so the duplicate case was never routable; the ordinals didn't break routing, they risked persisting a *synthetic* label into the transcript, which is the one artifact here that can't be regenerated. Colliding names are now blanked, suppressing their voice blocks while descriptions still render — matching what `_voice_notes` and the absorb stage already do with this case.

One refinement: the predicate is exact case-folded duplication, **not** `serialize.confusable`. `confusable` reports "Winifred Vance" beside "Winifred Vale" as colliding, but those are the headings that actually render and `match_name` resolves each exactly — only the bare "Winifred" is ambiguous, and nothing writes that. Blanking them would cost two distinguishable characters their voices to prevent nothing.

**Character pins** (P2) — real gap, fixed. `_CAST_SECTIONS` named only `character_state`/`transient_state`, so a pin promised something it didn't deliver. `voice_policy` stays out: it's `LOCK_IN` and needs no pin.

**Frozen fixture** (P1) — reverted, both copies. The coverage it was after now exists as an ordinary test (`test_an_anchor_reaches_the_assembled_prompt`), so the edit had stopped buying anything the rule doesn't already forbid. The fixture README records the attempt and the reversal, since the argument for it will recur.

**Editor hint** (P2) — fixed; it no longer claims "every turn".

**Not taken — the stale backlog** (P2). The count renders only in the `mode === "grid"` branch, and the only route back to it, `backToGrid`, calls `reload()`. So it can't be observed stale. Worth noting how that was established: I *did* implement a local `setChars` patch and a test for it first — then found the test passed with the fix removed, because `queryByText` returns null whether the count is absent-because-correct or absent-because-unrendered. The patch is gone as unearned and there's now a test pinning the round-trip property its absence relies on.

`check-mypy` still fails on `atomic.py`/`proclock.py` — untouched by this branch, reproduces with everything stashed, Windows-only. Deliberately not baselined.

@codex review

charlesmsiegel:
Round 2 addressed in `a5d94643e`. All four taken.

**Caps ran before macro expansion** (P2) — the sharpest of the four, and correct. `{{user}}` is eight characters that become the joined player names, so the sent text could exceed a ceiling documented as *the longest the prompt sees*. Expansion now happens first, with a test that fails against the old ordering. It does widen the generator/judge gap by one step since the judge reads the stored anchor — but substitution is already on that documented divergence list, and a cap that doesn't bound what is sent is the worse of the two.

**CRLF blank lines** (P2) — real, and it bit precisely the text most likely to have a boundary: a card imported or hand-edited on Windows. Newlines are normalised before the search. Writing the test also caught the half-cap floor working as designed — a boundary ten characters into a thirty-character cap is correctly rejected for the hard cut, so my first fixture was wrong, not the rule.

**Spec and plan still mandated the ordinals** (P2) — the one with the longest tail, since it would have told a later spec-driven change to restore labels round 1 removed. The spec now carries the blanking rule *and* both earlier mistakes, because each looked right at the time: `confusable` first (unsatisfiable by disambiguation, inverted argument convention), then the ordinals (synthetic label into an unregenerable transcript, buying nothing since the duplicate case was never routable). The plan's Task 2 is marked superseded rather than rewritten — a plan is a record of what was tried.

**templates/README.md** (P3) — it owns the input contract and listed three variables; a contributor following it would have omitted `correction` and either tripped `StrictUndefined` or silently stopped exercising the branch. Now documents it, including that the *caller* owns the provenance check.

Gate: `check-lint`, `check-eslint`, `check-templates` at baseline/passing; 348 backend tests green across the touched modules. `check-mypy` still fails on `atomic.py`/`proclock.py` — untouched here, reproduces with everything stashed, Windows-only, deliberately not baselined.

@codex review

charlesmsiegel:
Round 3 addressed in `54afbf4e3`. All four taken.

**NPC sharing a player's name** (P1) — right, and it's the wider version of the round-1 bug I only half-fixed. `voice_safe_names` blanked NPC-vs-NPC collisions but saw only NPC names, so an NPC named like the player still got a heading — and `split_reply` sends a player-labelled block to `speaker None` (*"never store a forged player line"*), persisting that NPC's dialogue as unattributed narration. Now tested with `serialize.match_name` against the present players, which is the same predicate `_speaker_and_role` applies, so this can't disagree with the persistence path about who a label names. `RESERVED_LABELS` too.

**Player pins holding NPC sections** (P2) — real. Sections drop whole, so pinning your own PC made every NPC anchor and example undroppable; on a large cast that alone can push a budgeted prompt over. Split into `_VOICE_CAST_SECTIONS`, applied only when a pinned actor is an NPC.

**Re-reading anchors in campaign listings** (P2) — also real, and my round-1 patch caused it by resolving through the overlay per row and discarding both scans. One directory scan now settles which rows even have a campaign-side file; only those are read. Worth recording that this took three attempts: the first called the world scan a *third* time, and the second assumed an inherited row can't carry a campaign anchor — but `set_voice_anchor` writes that file without copying the card down, so a tombstone can sit on an inherited character. The tombstone test caught it. There's now a test pinning the read count.

**`Nameless0` fixture label** (P1) — fixed, uses the approved placeholders.

1,251 tests green across the touched modules; `check-lint`, `check-eslint`, `check-templates` at baseline/passing.

Standing note, unchanged across all three rounds: `check-mypy` fails on `atomic.py`/`proclock.py`, untouched by this branch, reproducing with everything stashed. Windows-only mypy resolution, so CI does not see it, and it is deliberately not baselined — that would record a Windows-specific count and break the gate on Linux.

PR #425, opened 2026-08-25
Visual node/edge editor for greeting plot maps

### LINKED ISSUE #9: Visual node/edge editor for greeting plot maps
## Context
Greeting plot-map edges are stored and served, but edited only as flat chip-lists:
- `PUT /worlds/{wid}/greetings/{gid}/edges` — `backend/src/grimoire/routes.py:668` → `store.greetings.set_edges`. Body `Edges = {leads_to: [...], excludes: [...]}` (`routes.py:171`).
- `GET /worlds/{wid}/greetings/{gid}` returns `edges` (`{leads_to, excludes}`) and `predecessors` (`routes.py:644-654`).
- Client: `api.setEdges` (`frontend/src/api/client.ts:298`), `api.readGreeting` returns `GreetingDetail` with `edges` + `predecessors` (`client.ts:100`), `api.listGreetings` (`client.ts:288`).

Today `frontend/src/components/GreetingEditor.tsx` edits edges as two `chip` lists ("Leads to" `GreetingEditor.tsx:233`, "Excludes" `GreetingEditor.tsx:242`) toggled per-greeting (`toggle()`, `GreetingEditor.tsx:86`); the view side shows "Unlocks"/"Excludes"/"Depends on" as chip lists (`GreetingEditor.tsx:164-177`). There is no spatial/graph representation, so a multi-greeting plot is hard to reason about.

## Desired behavior
A visual graph editor for a world's greetings:
- Nodes = greetings (labelled by name); directed edges = `leads_to` (unlock) relationships; a distinct edge style for `excludes` (mutual-exclusion).
- Drag-to-connect (or click source then target) to create a `leads_to` edge; a control to flip an edge to `excludes`; delete an edge.
- Clicking a node opens that greeting (reuse the existing `GreetingEditor` detail/edit flow).
- Edits persist via `api.setEdges` for the affected source greeting; the graph reflects `predecessors` (reverse of `leads_to`).
- Coexists with the current chip-list editor (e.g. a "Graph" toggle on the Greetings tab) — the graph is an alternate view, not a replacement, so existing tests keep passing.

## Implementation notes
- New component `frontend/src/components/PlotMapEditor.tsx`, surfaced from the Greetings tab in `frontend/src/routes/WorldView.tsx:88` (toggle alongside `GreetingEditor`, or a sub-tab).
- Build edges from each greeting's `edges.leads_to` / `edges.excludes` (fetch per greeting via `api.readGreeting`, or extend the list endpoint — note `api.listGreetings` returns `Greeting` summaries without edges, so per-greeting reads are needed today).
- On edge create/delete, call `api.setEdges(wid, sourceGid, { leads_to, excludes })` with the full new arrays for that source (the route replaces, see `set_edges`).
- A lightweight hand-rolled SVG/absolute-positioned layout is acceptable to avoid a new graph dependency; if a library is added, keep it tree-shakeable.

## Acceptance criteria
- [ ] Greetings render as nodes; `leads_to` and `excludes` render as visually distinct edges.
- [ ] Creating an edge calls `api.setEdges` for the source greeting; deleting an edge removes it from that greeting's arrays.
- [ ] Clicking a node opens the greeting in the existing editor.
- [ ] The existing chip-list `GreetingEditor` flow remains reachable and its tests still pass.
- [ ] vitest test (`PlotMapEditor.test.tsx`) mocking the greeting/edges APIs: nodes render for greetings; adding an edge calls `setEdges` with the expected payload.

## Source
TODO.md §Frontend — deferred UI (not yet built)

charlesmsiegel:
Closes [#9](https://github.com/charlesmsiegel/grimoire/issues/9).

Greeting edges have always been authored as two flat chip lists per greeting — "Leads to" and "Excludes". That is the right shape to store and the wrong one to read: a plot of a dozen openings is a dozen chip lists that never say which one starts the story, what is downstream of what, or which branch you are looking at. Nothing in the app drew the thing those lists describe.

## What this adds

`frontend/src/components/PlotMapEditor.tsx` — the plot map as a graph, reached from a **List / Plot map** toggle on the Greetings section (world and campaign scope alike):

- A node per greeting, in columns by how deep in the unlock order it sits. A solid arrow per `leads_to`, a magenta dashed line per `excludes`.
- **Link ⇢** on a node arms it, then any other node is the far end; a chip picks which kind the new line is. Every other node is outlined while a link is armed, so the mode is visible rather than remembered.
- Clicking a line selects it: **Make exclusion**, or (on an exclusion) an unlock in either direction, plus **Delete link**.
- Clicking a node opens that greeting in the existing chip-list editor (the same `openGreeting` route a character page's greeting link uses).
- Campaign marks (played / done / skip) ride along on the nodes.

## It is a second view, not a second model

- Every write is `api.setEdges` for the greeting the line starts at — the same call the chip lists make — with that source's **whole** pair of arrays, because the route replaces rather than merges.
- Predecessors are derived, not fetched: an arrow into a node *is* the reverse of another node's `leads_to`.
- Nothing about the layout is persisted; it is recomputed on every read, so a map edited from either side agrees with the other and no new file joins the store.
- `GreetingEditor` is untouched, and List is what the section opens on.

## Whole-record writes, and the four rules that follow

The route **replaces** a greeting's arrays, so every write from this view is a whole-record write:

- A greeting whose edges never loaded may not be written at all — guessed-empty arrays would delete what the failed read did not see. Those nodes' link handles are disabled and their notice carries a **Retry**.
- A write applies to local state **before** it is sent, so a second edit to the same greeting builds on the first instead of racing it; each greeting's writes queue so its replacements land in the order they were made, and a failed write reverts that greeting.
- A scope change clears the map rather than leaving nodes on screen that would write the previous world's ids into the new one.
- Converting a mutual exclusion clears the far half first, so no intermediate state is an unlock coexisting with an exclusion that still blocks what it opens. A single two-endpoint route would be the real atomic fix and is worth its own issue.

## Decisions worth naming

- **Cycles are cut, then depths are relaxed.** A plot map is not guaranteed acyclic — nothing stops an author drawing a loop — and relaxation over one has no fixed point, so a DFS drops each back edge first. The arrow still draws; it just stops voting on which column its target belongs in.
- **Drawing over a pair that is already linked is refused, with the reason.** The alternative deletes an authored edge on one click; the line is on screen and selecting it can flip or delete it deliberately.
- **An exclusion recorded from both sides is one line**, because that is what it is (`store/greetings.py: availability` blocks in either direction) — so deleting it clears both halves. Converting it to an unlock asks which direction, since an exclusion is undirected and an unlock is not.
- **Hand-rolled SVG, no graph dependency**: columns of boxes and a bezier per edge. Several curve rules exist only because the map was driven in a real browser — a line that skips a column dips below the row (flat, it ran through the boxes between and landed exactly on top of their arrows), lines sharing a pair are lifted clear of each other so both stay selectable, and exclusions take `--alert` so a crossing is never ambiguous. Edge hit targets are a fat invisible twin under each 2px curve.

## Tests

- `PlotMapEditor.test.tsx` (21): nodes render for every greeting; the two edge kinds render distinctly; link / flip / delete call `setEdges` with the expected full payload; the refusal; a failed write leaves the drawn map alone; an unread greeting is a node but not a link source, and Retry recovers it; a gated `setEdges` proving the queued second write builds on the first; a `rerender` into a scope whose list has not answered; an exact canvas width for a three-greeting cycle; two ids containing spaces that a naive pair key would have merged; the keyboard edge-select leaving link mode; the empty map; campaign marks.
- One `WorldView.test.tsx` case for the toggle and the node → editor route.
- `make check-eslint` at baseline (no new findings, no baseline edit); `tsc -b` clean; the full vitest suite passes (124 files / 2555 tests). Backend is untouched.
- Also driven end-to-end in Chromium against an isolated store — seeding a six-greeting plot, drawing and deleting links, converting a mutual exclusion (which leaves the store with the unlock and both exclusion halves gone), and scrolling a map wider than the pane.


charlesmsiegel:
All nine review findings verified and fixed in c561bf4. Four of them turned out to be one fact seen from four sides — `PUT .../edges` **replaces** a greeting's arrays rather than merging, so every write from this view is a whole-record write — and that is now stated in the module docstring, since most of the fixes are consequences of it.

| Finding | Fix |
| --- | --- |
| P1 writes from greetings whose edges failed to load | Failed ids tracked; their link handle is disabled, `write` refuses them, and the notice carries a **Retry** that re-reads the scope |
| P1 concurrent whole-array writes race | The map the mutations read is a ref updated synchronously (so the second edit builds on the first), and each greeting has a promise chain so its replacements land in the order they were made; a failed write reverts that greeting |
| P1 stale nodes during a scope change | Greetings and edges are cleared with the scope, and the canvas waits for `ready` |
| P2 cycle depths unbounded | A DFS drops each back edge before the depth pass, so a loop is cut rather than merely stopped. A three-greeting cycle is three columns — asserted as an exact canvas width |
| P2 coincident edges unselectable | Lines sharing a pair are lifted clear of each other, so each curve (and its hit path) is distinct |
| P2 mutual-exclusion flip not atomic | See below |
| P2 arbitrary unlock direction | Both directions are offered as separate buttons, each naming its source |
| P2 keyboard select doesn't leave link mode | Pointer and keyboard now go through one `select` |
| P2 ambiguous pair key | Length-prefixed. `safe_id` does permit internal spaces — `natural_key`'s own docstring uses "SoL 2" — so this was reachable |

**On the atomic flip**: without a new endpoint there is no way to write two greetings in one operation, so instead the *ordering* was fixed so no intermediate state is a rule nobody wrote. The far half of a mutual exclusion is cleared first, and the unlock lands in the same write that clears the near half. Every intermediate is now either the exclusion the reader started with or the unlock they asked for — never an unlock coexisting with an exclusion that still blocks what it opens. A single two-endpoint route would be the real fix and is worth a follow-up issue rather than widening this PR.

Ten new tests cover the nine (a gated `setEdges` for the queued write, a `rerender` into an unanswered scope, an exact canvas width for a cycle, two ids with spaces). Full suite: 124 files / 2555 tests green; eslint ratchet at baseline. Also re-driven in a browser — converting a mutual exclusion to an unlock leaves the store with the unlock and both exclusion halves gone.

---

charlesmsiegel:
@codex review

---

charlesmsiegel:
Round 1's findings are fixed in 88a5d12 and 082c1c9 — the write queue is now per map rather than per greeting (the backend's `plotmap.json` read-modify-write is unlocked, so concurrent writes for *different* greetings race there), a failed write re-reads instead of reverting and discards what was queued behind it, a mutation half-done when the scope changes stops rather than finishing into the old scope, the chip-list editor is hidden rather than unmounted across the view switch so a draft survives it, a save that lands after the map opens makes it re-read, every greeting a mutation will write is checked before the first write, the edge toolbar is gone while the map re-reads, and a failed list request is no longer reported as an empty map.

@codex review

---

charlesmsiegel:
Round 2's six are fixed in 7c4c31e. Four were one shape — the chip list and the map both write a greeting's *whole* pair of arrays, and since the list editor now stays mounted behind the map, the two could hold different ideas of what those arrays are:

- Neither view writes while the other is mid-write (`onBusy` from the editor; the map refuses and says which view is holding it).
- A graph edit makes the mounted editor re-read the greeting it is **showing**; a greeting being **edited** is left alone, because a draft is the one thing on that screen that exists nowhere else.
- And a save whose chips were never touched no longer sends them at all — an untouched list is not the user's opinion about the plot map, which is also what makes leaving that draft alone safe.
- An exclusion with an unread endpoint cannot be converted in either direction: the old guard asked whether the far endpoint records the reverse exclusion, a question an unread endpoint answers "no" by construction.
- A reload queues behind writes already sent, and Retry is disabled while one is out.
- A reload that fails clears the map rather than re-enabling it over unconfirmed state.
- Both directions of one exclusion can no longer be converted at once.

One test was removed rather than repaired: "a second edit to one greeting builds on the first" drove a scenario the UI now refuses outright, so it could no longer fail for the right reason. What replaced it asserts the refusal — disabled controls, one request on the wire.

@codex review

---

charlesmsiegel:
**CI: `backend (py3.14)` failed on 082c1c9 — not this PR's, and here is why.**

The failure is `test_absorb_leaves_the_campaign_byte_identical`, and the entire difference is one second:

```
campaigns/run/activity.txt: b'2026-08-25T19:46:29Z\n'  !=  b'2026-08-25T19:46:28Z\n'
```

This PR's diff is frontend-only (`PlotMapEditor`, `GreetingEditor`, `WorldView`, `index.css`) — it does not touch `backend/` at all.

**Root cause.** The test snapshots the campaign tree, runs a full absorb, and asserts the tree is byte-identical afterwards, excluding only the scene's `.review.json`. But an absorb legitimately *writes* that durable sidecar, and every campaign-scoped write calls `campaigns.read.touch_quietly` → `_publish_stamp`, which rewrites `activity.txt` whenever the clock second has advanced (`if stamp <= read_activity(cid): return`). So the test passes only when the whole absorb finishes inside the same wall-clock second as its snapshot — which a loaded shared runner is exactly the thing to break.

**Reproduced locally on unchanged `main` code**, without any of this PR's changes:

- 1 of 3 bare runs failed — the cold one (4.7s vs 0.85s warm), which is the one slow enough to cross a second boundary.
- Deterministically: inserting `time.sleep(1.1)` between the snapshot and the absorb fails every time, with exactly the diff above. (That probe was reverted, not committed.)

**Proposed patch** (not applied here — it is a backend test change and this is a frontend PR, so it should land on its own):

```python
     after = {p: p.read_bytes() for p in croot.rglob("*")
-             if p.is_file() and p != review_file}
+             if p.is_file() and p not in (review_file, activity_file)}
```

with `activity_file = store.campaigns.paths.campaign_activity_path(cid)`, and the same exclusion in the post-cancel comparison. Naming the path rather than excluding a directory matches what that test already does for `.review.json`, and the reasoning is the same: the stamp moving is *correct* behaviour — `touch_quietly` fires from every campaign-scoped write, and the durable review sidecar is one — so it is not the #235 invariant this test guards. Freezing the clock for the test would work too, but the exclusion says what is true.

No re-run was needed: 7c4c31e was pushed before this event arrived, so CI is already running again on the current head. I will keep watching and drive it to green.

---

PR #443, opened 2026-08-28
Spec: read-path performance (cached derivations, conditional reads, SWR)

charlesmsiegel:
## What

Design spec for cutting data load time on the campaign/world list pages and opens: `docs/superpowers/specs/2026-08-28-read-path-performance-design.md`.

Traced causes: every navigation refetches `/api/shell`, which parses every `campaign.md` and the full transcript of every unabsorbed scene of the open campaign; `GET /campaigns` opens every scene file in the library; opening a world fetches full record lists once per index row and refires all of them on each section switch; and nothing is cached between requests on either side of the API.

The spec proposes four independently shippable layers, no store format changes:

1. **Extend `statcache`** to scene frontmatter heads, scene turn counts, campaign/world listing rows, and character-listing sidecar reads (sweep-sized memo families in their own pools).
2. **Intra-request duplicate-read elimination** — shell/campaign double parses, per-scene appearances-record re-reads.
3. **Conditional reads** — an in-memory library-wide read epoch (bumped from `revision.bump` plus a coarse mutating-2xx middleware trigger), spent as weak ETags/304s on an explicit allowlist of hot GET routes.
4. **Frontend stale-while-revalidate** payload cache for CampaignsView / WorldsView / WorldView / CampaignHub, invalidated over the existing `appEvents` bus; WorldView refreshes only the departed section's count.

## Status

Spec only — implementation follows the plan-gate pipeline. The Codex plugin isn't available in this remote session, so the spec→plan adversarial gate ran as a fresh-context adversarial review agent instead; findings will be addressed in follow-up commits on this branch. Recommend also running `/codex:adversarial-review` locally before implementation starts.

https://claude.ai/code/session_01VWzxXQoDZ2zgxusonw5Ck7

---
_Generated by [Claude Code](https://claude.ai/code/session_01VWzxXQoDZ2zgxusonw5Ck7)_


charlesmsiegel:
Stopping the spec-revision loop here. Six review rounds (22 findings) are folded into the spec; round seven's four P2 threads are left **open on purpose** as inputs to the implementation plan — each is real, none contradicts the design, and all four are implementable inside mechanisms the spec already defines:

- **Chronicle / campaign-changes aggregates**: give them layer-1 stat-keyed memos or take them out of the hot revalidation block — decided in the plan where those routes are touched.
- **Store-side bump hooks vs per-app epoch**: `revision.bump` and the usage append run in store code with no app handle; the plan resolves this with a registry of live app epochs bumped together (over-invalidation, the allowed direction), preserving per-app-instance minting.
- **Cross-tab store-root change**: the payload cache needs a cross-tab signal (or store-identity check) so another tab never paints the previous library — the spec's identity invariant extended to multi-tab.
- **Authoritative 404/410 evicts** the cached payload instead of leaving a deleted record's frame rentable.

Each fix-round has drawn a fresh batch of same-severity findings against a document that only catalogues them, so further polish belongs to the plan, where these become concrete code with tests.

---

Issue #227, opened 2026-07-14
Mechanics Phase 8 follow-ups: deferred minors from review


charlesmsiegel:
Batched non-blocking minors from the Phase 8 (module authoring UI) review waves, triaged follow-up by the final whole-branch review:

- module_edit.recover()/_replay_journal: defensive guard so an I/O error during replay can't break _apply's result-dict contract or abort startup lifespan (T2)
- upsert_content silently drops non-string metadata field values — comment or reject (T4)
- _migrate_file sheet_type branch reassigns data["fields"] (injects {} for malformed sheets); stray .json.tmp leak on crash between write and replace; missing trailing newline vs _write_json (T7)
- no test for world-sheets-migrate-regardless-of-current-binding (code correct) (T7)
- _impact walks _sheet_files twice when a migration is set — mergeable into one pass (T8)
- instantiate route maps SheetConflict to 400, other sheet routes use 409 — consider uniformity (T10)
- test_proposal_route_sites_locked is a near-tautological whole-module substring check — strengthen or drop (T11)
- upsert_sheet_type narrowing its groups list leaves a stray {"group": gid} layout node for that type (non-fatal display drop) (T3)
- client layout preview depth cap bounds only use-chains, not raw row/column nesting (backend caps all structural depth) (T18)
- importModule duplicates the ok-check/ApiError block (3rd copy) — dedup helper (T13)
- metaFromEntry String(v) is lossy for non-string frontmatter values — comment (T17)
- four-dash-fence frontmatter parse edge in the new empty-meta special case (T6)
- duplicate_module copies dotfiles/OS cruft verbatim (T9)

Spec: docs/superpowers/specs/2026-07-13-mechanics-phase8-authoring-ui-design.md
Plan: docs/superpowers/plans/2026-07-14-mechanics-phase8-authoring-ui.md

PR #445, opened 2026-09-05
Mechanics & entity fields: Phase 8 follow-ups (#227) and number/choice widgets (#221)

### LINKED ISSUE #221: Per-kind game-mechanical fields on entities
Deferred from the entity-kinds design (docs/superpowers/specs/2026-07-11-entity-kinds-design.md), split out of #37.

v1 typed fields are minimal text scalars (item_type/rarity, group_type, creature_type/threat). Full stat blocks for creatures, item mechanics, and group resources belong to a later pass that hooks into the Mechanics & Dice roadmap's Phase 1 data contract, so mechanical fields are defined once, not per-feature.

Depends on #36/#37 (kinds + field registry) and the Mechanics & Dice data contract.

charlesmsiegel:
Resolves #221 and #227.

## #227 — Phase 8 review follow-ups

All thirteen deferred minors, each with a test:

- **T2** `recover()` survives an I/O error during journal replay. The journal stays for the next attempt, the debris sweep is skipped as for a quarantined one, and `recover()` now returns the modules whose journal is stuck: `_apply` and `delete_module` refuse those, because journals replay in nonce order (random) and a second journal written behind a stuck one could replay first. A journal that parses but whose migration lacks what its op reads is quarantined rather than raising `KeyError` out of startup.
- **T4** `upsert_content` refuses a non-string metadata value instead of dropping it.
- **T7** `_migrate_file` reassigns `fields` only from the ops that edit it (a type rename no longer replaces a malformed value with `{}`), and writes through the sheet writer's own function so a migrated sheet is byte-for-byte what a save writes. `store.atomic` already discards its temp on failure, which covers the `.tmp` leak concern. Plus a test that world starting sheets migrate regardless of any binding.
- **T8** `_impact` counts migrations in the same walk that judges validity.
- **T10** Both instantiate routes answer 409 for a sheet CAS conflict, through one shared mapping.
- **T11** The proposal-site lock check walks the AST: every `proposals.new` call must sit lexically inside a `campaign_lock` block.
- **T3** Narrowing a sheet type's groups prunes that type's `{"group"}` node and the group's names from its layout, scoped so other composing types keep theirs. The prune skips a non-string group value rather than raising on it.
- **T18** The client layout preview's depth cap counts row/column nesting, not only use-chains, mirroring the server's `MAX_DEPTH`.
- **T13** The two zip-import calls share one reply parser with `requestForm`.
- **T17** `metaFromEntry` says why `String(v)` is a narrowing, not a lossy conversion.
- **T6** A closing fence is a line that is exactly `---`, in the general path and in the head parser alike, so a list endpoint and a full read cannot describe one file differently.
- **T9** `duplicate_module` and `export_module` leave dotfiles and OS cruft behind, through one shared pack-member predicate.

## #221 — entity fields declare their own constraints

The mechanical half of #221 (full stat blocks for creatures, items and groups, defined once by the module) landed as sheets in Phase 3. What the re-scoping comment left open was field richness on the descriptor itself: widgets beyond text, and validation both sides can see.

Two widgets join `text` and `ref`. A `number` carries optional inclusive `min`/`max`; a `choice` carries either a literal `options` tuple or a named `source` in `OPTION_SOURCES` for lists only known at runtime. The spec is the constraint: `entity_schema.valid_value` refuses what a spec does not admit, and `EntityEditor` renders the control the spec implies. The bespoke `VALIDATORS` table goes.

- `climate` is a choice sourced from the climates list, read fresh so a climate saved a moment ago is offered at once. A stored value none of the options answers to stays visible and clearable in the picker rather than being blanked, and is called "not an option" only once the list has actually arrived. The sidebar names a known option by its label.
- `persistence` is a number bounded 0..1.
- A `text` value must be a string, the rule `upsert_content` applies to module content.

No new per-kind field is invented: the flavour fields stay free text, and which of them should become a scale is the world author's call. Sheets remain the place for mechanics; the descriptor docstring says why.

## Gates

Rebased onto `main` after #446 (the mypy baseline fix) merged; the copy of that commit this branch carried was dropped as already upstream. `make check` targets run locally: `check-lint`, `check-mypy`, `check-eslint`, `check-templates` and `check-web` pass. `check-py` (coverage 93.26%) and `check-pydantic1` each pass every test but `test_a_read_only_record_is_not_silently_replaced`, which asserts that a `0o444` file refuses a write; this container runs as root, which the kernel exempts from file modes, so the refusal cannot be observed here. CI runs unprivileged and passes it. Two baselines shrink: mypy's `layout.py` entry (typing the prune parameters as `AbstractSet` resolved it) and three eslint counts on `client.ts` (the shared reply parser).

The Codex review skills named in CLAUDE.md are not installed in this environment; the built-in `/code-review` was run over the branch at high effort instead, and its eight findings are addressed in the "Review fixes" commit. The Codex reviewer bot on GitHub reported its usage limit is exhausted, so it has not reviewed this PR.

https://claude.ai/code/session_01J9fMmjrskZKRcbWexF9Q6J


charlesmsiegel:
CI's `mypy` job failed on this PR's first head with three rows this branch never touches (`characters.py` arg-type 4→5, `chronicle.py` return 0→1, `facts.py` return 2→4). The same job is red on `main` itself, for its last three runs including the current head `194d610`: the ratchet baseline was last regenerated on 2026-08-28 and five findings landed after it.

The fix is [#446](https://github.com/charlesmsiegel/grimoire/pull/446), which annotates `_ProcessScopedLock.__exit__` as `Literal[False]` (mypy read the `-> bool` as "may swallow", and reported every function that returns only inside `with campaign_lock(cid):` as missing a return) and narrows `into_cid` in `import_card`'s `update_vid` branch. This PR now carries that same commit so it can go green; the port no-ops once #446 merges. The PR description's earlier "Python 3.11 vs 3.12" explanation for these rows was wrong and is superseded by this.

---

PR #447, opened 2026-09-23
Cut latency and load times at large-library scale (web + Android)

charlesmsiegel:
## What this does

This PR goes after work on hot paths whose cost grows with the whole library — every world, campaign, scene, image and ledger row — rather than with what the screen shows. Everything was profiled against a large synthetic store that uses placeholder names only. Each change was measured before and after, reviewed adversarially on its own, and then the whole branch was reviewed once more across units (below).

**Status:** complete. All units and the final whole-branch review fixes are merged on this branch.

### Integrated before/after (base `9afe175` → this branch)
Production builds, each served by its own backend on a hardlink clone of the synthetic store, with the store resolved through the bootstrap pointer (the production path). Interleaved medians in Chromium; 4× means 4× CPU throttle, used as the Android proxy.

| | Base | This branch |
|---|---|---|
| World: Characters click → first card (1× / 4×) | 388 / 609 ms | **55 / 159 ms** |
| World: cold deep link → first card (1× / 4×) | 957 / 1407 ms, 55.8 MB | **169 / 624 ms**, 1.5 MB |
| World: revisit A→B→A, A's first card (1× / 4×) | 806 / 957 ms | **16 / 92 ms** |
| World overview: API requests on open | 79 | **14** |
| Campaign hub, cold, complete with counts | ~6.6 s | **~0.6 s** |
| Long scene, cold deep link, settled | ~8.1 s | **~0.55 s** |
| Chat turn: pre-LLM / post-LLM | 1402 / 503 ms | **587 / 375 ms** |
| `/api/shell` warm | 652 ms | **112 ms** |
| World characters warm (world / campaign scope) | 46 / 186 ms | **7 / 14 ms** |
| Scene open warm (windowed) | 61 ms | **8 ms** |
| Global keyword search warm | 5.1 s | **1.5 s** |
| Startup lifespan (identity backfill) | 185 ms (160) | **25 ms (1)** |
| Android sim: port reported after spawn | 989 ms | **24 ms** |
| Android sim: keep-alive request | ~52 ms | **~8 ms** |
| Android sim (4×): cold world grid first card | 1660 ms, 43 MB | **1020 ms**, 1.6 MB |
| Android relaunch in the same WebView profile | 0 cached, 38.7 MB | **31 cached, 57 KB** |
| Desktop launcher as shipped: cold `/` first card | 995 ms (Vite dev) | **316 ms** (built bundle) |

### Opening a world and its character list (the priority)
- **Character list endpoint.** A per-character row memo stamped on every file and directory the row reads. ctime catches in-place edits and restored-mtime rewrites. Listings use `scandir`, and the JSON encoding is cheaper. Response bodies are byte-identical.
- **Fewer competing requests.**
  - `GET /worlds/{wid}` carries the campaign count, or `null` if any campaign is unreadable, so the page stays up.
  - The describe queue has a `?count=1` form.
  - Module reads wait for the section that needs them.
  - Section clicks stop re-downloading every list just to count it.
  - Plot-map edges ride the greeting list, so the overview and the plot map no longer fire one request per greeting.
- **The grid paints from memory, on intent.**
  - Remembered rows are keyed by store root and scope, forgotten on any write or data-dir change (including a move made in another tab), and always revalidated.
  - Intent prefetch starts on hover, focus or pointerdown; on touch it waits for a resting finger, not a scroll.
  - The first screenful draws first, from memoized cards.
- **Thumbnails, not originals.** Grids, rails and portraits request snapped `?w=` sizes with `srcset`/`sizes`, loaded lazily; a location slot wider than 1024 can still pick the original. The thumbnail pipeline:
  - draws a thumbnail as the browser draws the original: EXIF orientation only where Chromium applies it, the ICC profile kept, animated sources served as originals;
  - uses WebP, or JPEG/PNG where Pillow has no WebP encoder, **as on the Android build**;
  - uses a store-relative cache key, so a synced or moved library reuses its thumbnails;
  - is single-flight, and downscales before converting;
  - carries a thumbnail ETag and URL keyed on the pipeline generation, so cached copies turn over when it changes;
  - sweeps only older generations, so mixed-version devices on a synced library leave each other's caches alone.

### Backend
- **Scene open** stops parsing the campaign-wide response ledger under the campaign lock. The ledger is written as compact JSON, which older builds still read.
- **Pointer memo and overlay `View`.** `paths.home()` stops re-reading `~/.grimoire.json` on every call. Overlay listings resolve roots and tombstones once per sweep instead of once per row.
- **Id/name roster** for callers that only counted full character rows: shell, todo, cast-changes, suggestions, the turn's cast directory and sheet coverage.
- **Turn loop CPU.**
  - No discarded work in an actor-scoped compose.
  - Incremental stream watchers (previously O(n²) on the event loop).
  - Datetime macros resolved once.
  - Header-only history reads.
  - Per-compose token-count memo.
  - Throttled heartbeats.
  - Prompts are byte-identical.
- **Usage and search.** Campaign-prefiltered ledger parsing, a memoized unpriced scan, a search memo pool sized above the corpus, and one query in flight per mode.
- **Startup and background I/O.**
  - The identity-backfill watermark makes startup O(campaigns) instead of O(scenes).
  - `best_stamp` no longer calls `strptime`.
  - A tiktoken load failure is remembered.
  - The SSL context is built lazily.
  - Backups store media uncompressed and sweep stale temps.

### Frontend, delivery, Android
- **Play view.** Memoized transcript rows, a streaming markdown view parsed a block at a time, a gated CastPanel, a memoized inspector and columns. At 600 posts, a keystroke costs about 1–4 ms instead of about 90 ms.
- **Navigation.** One `/api/shell` per arrival. The hub and scene list no longer wait on it, and they never draw another campaign's payload or an unread price as `$0.00`. One `/api/config` at boot.
- **Serving.** No gzip for loopback clients that send no forwarding header. `immutable` on hashed `/assets/*`, `no-cache` on `index.html`.
- **Android entry.**
  - `IPPROTO_TCP` plus `TCP_NODELAY`.
  - The port is bound and reported before imports.
  - One app build.
  - Routers are composed by reference on FastAPI 0.115. Where FastAPI analyses routes lazily (the desktop set), one include branch per domain is kept, so the first request after a restart is not slower.
  - No access log.
  - A stable port per install, so the WebView's caches and localStorage survive restarts.
  - The service starts before the WebView.
- **Desktop launcher.** Serves the built bundle from the backend on `:8173` and rebuilds it when stale. If a rebuild fails, the browser is told it is getting the previous build. `--dev` / `-Dev` keeps Vite and HMR on `:5173`. The README explains the address change and which localStorage conveniences start over.

### Deliberately not done
- **Splitting `responses.json` into per-scene files.** This is the largest remaining per-turn cost on a long campaign: the ledger is still rewritten several times per turn. It is a one-way store-format change on a library that can sync between devices, so it needs an explicit decision.
- **Route-level code splitting.** Measured neutral to negative when the bundle is served from the same machine.
- **A 1-wide read gate against GIL convoys.** It changes request scheduling globally.
- **A persisted (localStorage) grid cache for the very first paint.** A cold deep link is dominated by boot and config, not the list round trip.

### Known follow-ups
- **Campaign copy of a world.** Still lists every record kind in full to draw the column counts. Overlay-aware counts need their own endpoint.
- **Character page.** Opening one from the roster is unchanged: 12 requests, including the whole roster, lore and greeting lists.
- **Android import failure.** An import failure after the early port report shows the WebView's error page instead of the loading screen's text. The message still reaches logcat.
- **Boot contention and palette prefetch.** `/api/shell` at boot still overlaps a cold deep link's list read, and the command palette's world items do not prefetch.

## Verification
- **Per unit.** Each unit ran:
  - its targeted tests and guards;
  - the full backend suite, plus vitest where relevant;
  - the ruff, mypy, eslint and template ratchets;
  - its version-sensitive tests under the pydantic 1.10 / FastAPI 0.115 set.

  A separate reviewer then re-measured the unit and fixed what it found.
- **Final whole-branch review.** Four lenses — cross-unit caching, conventions and platforms, request coverage and UX, and integrated measurement — produced findings. They were fixed in 13 commits, each with a test that fails without it, and an independent verifier checked the fixes. That review found two things per-unit review could not:
  - the Android Pillow wheel has no WebP encoder, so thumbnails there fell back to originals;
  - thumbnails ignored EXIF orientation.

  Both are fixed.
- **Final head.**
  - ruff, mypy and templates are at baseline; the ruff/mypy/eslint baselines only shrank.
  - `tsc -b` is clean and vitest passes 3240/3240.
  - `vite build` succeeds.
  - The full backend suite passed on every integration apart from the two tests below.
  - CI's Android APK build is green.
- **Red CI.** The red backend and pydantic-1 jobs are **not this PR's**. The only failure is `test_frozen_campaign`, whose snapshot predates the greeting phase fields; `main` is red on the same test (the comment below has the one-command fix). `test_atomic`'s read-only test fails locally only because the container runs as root. A timing race in `test_world_fork` that CI hit was made deterministic (test-only).
- **Codex gates.** The repo's Codex review gates (`/codex:review`, `/codex:adversarial-review`) are not available in this environment. The adversarial review passes above stood in for them. Running the Codex gates before merge is recommended.

https://claude.ai/code/session_01DsuSMm3bVaDG2nvUg1eFJw


charlesmsiegel:
**`backend (py3.14)` is red; `backend (py3.11)` and `pydantic 1.10` will follow for the same reason. It is not this PR's failure.**

The single failing test is `test_frozen_campaign.py::test_the_frozen_campaign_still_reads_the_way_it_was_recorded` (1 failed, 9085 passed). `main` fails the same test in the same way on every run since `2a9e728` ("Add bounded greeting phase recommendations", CI runs 964–967). That commit added three fields to greeting reads, `optional`, `phase` and `sequence`, without regenerating `backend/tests/fixtures/frozen_campaign/snapshot.json`.

The failure is deterministic, so a re-run cannot change it and I have not spent one.

Two checks I ran:
- At the base commit (`9afe175`), regenerating the snapshot per the fixture README changes exactly 18 lines: `"optional": false`, `"phase": ""` and `"sequence": null` on each of the six greeting entries. Nothing else moves.
- The same regeneration run on this PR's head produces a **byte-identical** snapshot, so none of the changes here alter how the frozen campaign reads back.

**Proposed fix** (belongs with the greeting-phase change, not this PR):
```
cd backend && PYTHONPATH=src .venv/bin/python -m tests.fixtures.frozen_campaign.sweep
git add tests/fixtures/frozen_campaign/snapshot.json
```
I have not widened this PR with it. Say so if you'd like it carried here instead.

---

charlesmsiegel:
@codex

